PR #2, opened 2020-08-29
overlay: use OverlayPanel instead of Overlay

melkypie:
Allows resizing and using the Runelite's colorOverlayBackground config option to recolor the background of the quest helper overlay.


Zoinkwiz (approved):
Looks good, thanks 😄.

PR #4, opened 2020-09-05
Add Underground Pass guide, and bump runelite version to 1.6.26.1

llamositopia:
I'm back, with the UndergroundPass guide completed.

For the doll-imbuing portion of the quest, I had to rely on some varbit stuff, so I stuck a todo on each to confirm that it's the right varbit.

I also bumped the version number on the runelite dependency to match the current stable release, as of writing this.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ repositories {
 	mavenCentral()
 }
 
-def runeLiteVersion = '1.6.25'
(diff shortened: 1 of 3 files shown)


Zoinkwiz (approved):
Looks good, I'll have to try this out once I have an account ready to do Underground Pass!

PR #47, opened 2020-11-04
Updated tail of two cats to save 5 min and 20 sec

Haavardaw:
changed it to skip the cutscene option instead of wasting time


Zoinkwiz:
I'm tempted more for this to remove a highlight option entirely, I think if someone is using this helper to just help with quests vs speed through them they shouldn't be accidentally skipping meaningful content. Equally shouldn't force people to watch the cutscene.

Zoinkwiz:
I've removed it for now in https://github.com/Zoinkwiz/quest-helper/commit/b437c9c091fd867ae0046b174e017b48dce5d5da, so closing this PR.

PR #48, opened 2020-11-04
DesertTreasure Ask about the Diamonds of Azzanadra "yes" option

Haavardaw:
Did a fix to this, should hopefully work


Zoinkwiz:
Unfortunately this wouldn't work as the 'Yes.' option is the first option in both the incorrect and correct cases. I've implemented a new functionality which allows you to exclude highlights based on other present text with https://github.com/Zoinkwiz/quest-helper/commit/1605fbb3279c607545024d27b5c11f897d20e92d, which should also work for other similar scenarios which we'll need to re-visit.

PR #49, opened 2020-11-05
Updated coordinate on Leenz

Ghordrin:
Updated the coords for Leenz. Wrong coordinates on X-axis.


Zoinkwiz (approved):
Looks good, thanks!

PR #57, opened 2020-11-10
Fix checks for collecting beard tax from Keepa and Flosi

### LINKED ISSUE #43: Missing dialog during fremennik isles
During the part of the quest where u collect taxes from bearded people, it's missing the dialog with Keepa Kettilon in the fishing shop.

andmcadams:
Add collectedKeepa to the conditions for collecting from Flosi and talking to Gjuki.
Should fix #43 but needs testing.


Zoinkwiz (approved):
Looks good, good spot, thanks for fixing 😄

PR #64, opened 2020-11-13
Added Holy Grail Quest

kijjuy:
Ran through quest on 2 separate accounts, worked on both.

### DIFF
--- a/src/main/java/com/questhelper/quests/holygrail/HolyGrail.java
+++ b/src/main/java/com/questhelper/quests/holygrail/HolyGrail.java
@@ -1,3 +1,27 @@
+/*
+ * Copyright (c) 2020, Kijjuy
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
 package com.questhelper.quests.holygrail;
 
 import com.questhelper.QuestHelperQuest;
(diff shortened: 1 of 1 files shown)


src/main/java/com/questhelper/quests/holygrail/HolyGrail.java
```diff
@@ -0,0 +1,362 @@
...
+		highlightGrailBell.setHighlightInInventory(true);
+	}
+
+	public void loadZones()
+	{
+		camelotGround = new Zone(new WorldPoint(2744, 3517, 0), new WorldPoint(2733, 3483, 0));
+		camelotUpstairsZone1 = new Zone(new WorldPoint(2768, 3517, 1), new WorldPoint(2757, 3506, 1)); //og. 2764, 3517. 2748, 3501
+		camelotUpstairsZone2 = new Zone(new WorldPoint(2764, 3517, 1), new WorldPoint(2748, 3496, 1));
+		merlinRoom = new Zone(new WorldPoint(2768, 3505, 1), new WorldPoint(2765, 3496, 1));
+		entranaBoat = new Zone(new WorldPoint(2841,3332,1), new WorldPoint(2823,3328,1));
```
Zoinkwiz, line 175: I'd extend this + the entrana Island to range from planes 0-2. Although we don't expect a player to change plane, it's possible and an easy thing to catch with a single digit change.

src/main/java/com/questhelper/quests/holygrail/HolyGrail.java
```diff
@@ -0,0 +1,362 @@
...
+		highlightMagicWhistle2.setHighlightInInventory(true);
+
+		highlightGrailBell = new ItemRequirement("Grail Bell", ItemID.GRAIL_BELL);
+		highlightGrailBell.setHighlightInInventory(true);
+	}
+
+	public void loadZones()
+	{
+		camelotGround = new Zone(new WorldPoint(2744, 3517, 0), new WorldPoint(2733, 3483, 0));
+		camelotUpstairsZone1 = new Zone(new WorldPoint(2768, 3517, 1), new WorldPoint(2757, 3506, 1)); //og. 2764, 3517. 2748, 3501
```
Zoinkwiz, line 172: What's the `og.` about?

src/main/java/com/questhelper/quests/holygrail/HolyGrail.java
```diff
@@ -0,0 +1,362 @@
...
+
+		findFisherKing = new ConditionalStep(this, goToGalahad);
+		findFisherKing.addStep(inFisherKingCastle1SecondFloor, talkToFisherKing);
+		findFisherKing.addStep(inFisherKingCastle1BottomFloor, goUpStairsBrokenCastle);
+		findFisherKing.addStep(new Conditions(hasGrailBell, inFisherKingRealmAfterTitan), ringBell);
+		findFisherKing.addStep(talkedToFisherman, pickupBell);
+		findFisherKing.addStep(inFisherKingRealmAfterTitan, talkToFisherman);
+		findFisherKing.addStep(new Conditions(hasExcalibur, titanNearby), attackTitan);
+		findFisherKing.addStep(new Conditions(hasTwoWhistles, inTeleportLocation, hasExcalibur), blowWhistle1);
+		findFisherKing.addStep(new Conditions(hasTwoWhistles, hasExcalibur), goToTeleportLocation1);
```
Zoinkwiz, line 110: This looks like a player will get stuck here if they haven't taken excalabur with them to Draynor Manor to get the whistles. If we're wanting require the player to get Excalabur before this prompt, probably worth us adding in a 'go get excalabur from the bank' step between this and the below, like:

`findFisherKing.addStep(new Conditions(hasTwoWhistles), goGetExcalabur);`

Zoinkwiz (changes requested):
This looks really good and clean, thanks! Left a couple comments but should be good after.

Could you also please use the formatter in https://github.com/runelite/runelite/wiki/Code-Conventions to clean up the couple formatting variations then push those changes? Thanks 😄

src/main/java/com/questhelper/quests/holygrail/HolyGrail.java
kijjuy: are you including camelotUpstraisZone1 and 2 in expanding the planes? Because that is meant to be a specific region on the same floor as Merlin.  Or are you specifically talking about expanding the planes of the entranaBoat and entranaIsland locations?

kijjuy:
I'm pretty sure that I am using the RuneLite formatting tools. Both the code style setting in IntelliJ and the CheckStyle plugin say that I have everything formatted properly. But it is also the first time for me using both of these things, so it is very possible that I messed something up somewhere and it's giving me a false positive result.

src/main/java/com/questhelper/quests/holygrail/HolyGrail.java
Zoinkwiz: Specifically the entrana ones.

kijjuy:
I updated the quest to implement the comments that you left. Sorry if I'm spamming you with notif's from GitHub, I'm not really sure if I'm supposed to leave a comment when I update my branch or just kinda update it and wait for you to see it.

Zoinkwiz (approved):
Looks good, thanks!

PR #71, opened 2020-11-16
plugin: inject members immediately after decleration

zeruth:
This resolves issues with an EventBus that subscribes automatically rather than using a register() function.
(it was possible to receive events before the plugin was ready for them, causing havoc.)


Zoinkwiz (approved):
Looks good, thanks 😃

PR #75, opened 2020-11-18
Logs for fires ItemCollections

Haavardaw:
Added all logs that are able to be firemaked for quest reqs like "any log" if its needed for fire


Zoinkwiz (approved):
LGTM, thanks!

PR #77, opened 2020-11-18
Add Clocktower

Zoinkwiz:


src/main/java/com/questhelper/quests/clocktower/ClockTower.java
```diff
@@ -0,0 +1,233 @@
...
+
+    @Override
+    public ArrayList<ItemRequirement> getItemRequirements()
+    {
+        ArrayList<ItemRequirement> reqs = new ArrayList<>();
+        reqs.add(bucketOfWater);
+        return reqs;
+    }
+
+//    @Override
```
Poltergeisen, line 198: Consider removing dead code?

BenjaminLee-boop (approved):

Zoinkwiz:
Updated + made actually work in https://github.com/Zoinkwiz/quest-helper/pull/209

PR #79, opened 2020-11-20
Fixed issue where widgets would fire 2 times in ConditionalStep

Patyfatycake:
Since QuestStep already subscribes to the "onWidgetLoaded" event we want to override and re-use that subscription instead of creating another one. Hence ConditionalStep should override not subscribe.


Zoinkwiz (approved):
Good spot, thanks 😄

PR #80, opened 2020-11-20
Demon slayer - hightlight incantation

Patyfatycake:
Requires this PR before it works https://github.com/Zoinkwiz/quest-helper/pull/79

Quest


Zoinkwiz (approved):
Looks great, thanks! I can try adding this to Shadow of the Storm as well, I think the same structure should work perfectly 😄

PR #82, opened 2020-11-21
Fix coordinates in The Hand in the Sand

Kebabpizza:
* Sandy's desk wasn't being highlighted
* The doorway indicator was one tile too far to the west 
bild


Zoinkwiz (approved):
Looks good, thanks

PR #90, opened 2020-11-24
Add The Giant Dwarf

slaytostay:
Currently only gives the option to join the opal company. I'm not sure what would be a good way to track which company tasks need to done, have been done or have been completed.


Zoinkwiz (approved):
Looks really good, definitely seems an annoying quests due to the lack of states!

PR #95, opened 2020-11-28
Added Recruitment drive

Patyfatycake:


Zoinkwiz (approved):
Really good stuff, handles the various states and actions very well 😄 . We should add a proper reset for when you fail puzzles, but we can do that at a later date.

PR #98, opened 2020-11-30
Added Dwarf Cannon

julysfire:
Addition of the Dwarf Cannon quest.


Zoinkwiz:
Rebased and fixed a few things, and merged in with https://github.com/Zoinkwiz/quest-helper/pull/99. Main things were:

- Use substeps to keep the sidebar cleaner. Rather than 6 'fix railings', you can just have 1 which they're all grouped under.
- Objects which change in any way are usualyl represented as a NullObjectID instead. You can check the linked PR for examples.

PR #101, opened 2020-12-06
Add quest Fight Arena

Zantier:
Needs testing


Zoinkwiz (approved):
Looks really good, thanks! I added in a couple talking steps to catch cases where people may click off the dialog. Wasn't able to find a way to get those skipped steps, but good we're catching them anyway.

PR #108, opened 2020-12-09
Add The Feud quest

Patyfatycake:


julysfire:
Looks like there is a break between getting Drunken Ali drunk with 3 bears and then picks back up again with talking to Ali The Camel man to get the two receipts.

Patyfatycake:
Think I resolved the 2nd one. Although the first and 3rd one should be fine? I can't see anything immediately in the code. Also don't have anything to test it sorry.

Zoinkwiz (approved):
Looks good, thanks 👍

PR #119, opened 2020-12-19
Added Devious Minds quest.

julysfire:
Let me know if anything needs to be changed :)


julysfire:
Going to withrdaw so I can make some changes (GeneralRequirements for example) and to reduce the commit to just the Devious Minds quest (Haven't really used git before this so been a learning experience).

PR #128, opened 2021-01-10
Small fix in A Tail of Two Cats

julysfire:
The Catspeak amulet needs to be equipped as both steps that use this required it equipped as you talk to your cat and the NPC at the same time.


Zoinkwiz (approved):
LGTM, thanks

PR #133, opened 2021-01-12
Create a shareable run configuration for IntelliJ 2020.1+

Senmori:
This will work for IntelliJ versions 2020.1+.
Multiple run configurations can be added to the .run/ directory. (i.e. build configurations, etc.)

The configuration name can be changed of course. As for the JDK; I personally use OpenJDK 13 but used the Bundled option so it should work with any installation of IntelliJ.

This should make it extremely easy for people to start doing some dev work for this project. If they don't have much experience with Java or IntelliJ it can be a bit of a pain to figure out how to run Runelite via this project. This allows them to not have to worry about it and just start the actual coding.

Picture of Run Configuration: https://i.imgur.com/h59n63u.png


Zoinkwiz (approved):
Looks good, thanks for this 😄

PR #136, opened 2021-01-13
Add missing Requirements

Senmori:
Lots of changes in this one. This PR adds missing `Requirement`s as well as removing some duplicate code.

I'm marking this as a draft because it's 3AM and I wanted

Quest Requirements are found at: https://oldschool.runescape.wiki/w/Quests/Requirements

I was able to run QuestHelper without any errors with these changes but I did not test quests that specifically used some of these requirements.

I would appreciate it if someone could review this and see if I missed anything or changed some logic somewhere.

****1. Duplicate code****
Duplicate code was found in several classes, mainly `getDisplayTextWithChecks` where all it did was use the same `LineComponent` builder to combine the display text and `getColor`. This bit of code was moved back into the abstract `Requirement`.
`getColor` was also moved back into `Requirement` for code that just called `check(Client)` to determine if the color should be RED or GREEN.
Everything more complicated than `getColor(client) ? Color.GREEN : Color.RED;` was left alone.
Same thing for `getDisplayTextWithChecks`. If it was doing anything other than combining the display text with the Color it was left alone.

****2. Operation****
`Operation` is a simple enum that is used to calculate if a number is greater than, less than, equal to, or not equal to another number.
`Operation` now has a `BiFunction` attached to it. This allows for de-duplication of code, as well as it is just easier to read and maintain.
See `WeightRequirement` or `SkillCondition` for an example of how this is easier.

*****3. Missing Requirements****
`CombatLevelRequirement` checks if the client has a required combat level. There could be additional functionality where we could check if the combat level is anything other than `Operation#GREATER_THAN`, however there is no quest requirement that uses it.

`FavourRequirement` checks if the client has at least a certain percentage of favour with a certain house (i.e. Shayzien, Arceuus, etc).

`FreeInventorySlotRequirement` checks if the client has a certain number of `InventoryID` slots not in use. While this is primarily going to be used for the inventory, I did not specify the inventory just in case future inventories are added that could be used (i.e. Looting Bag)

`QuestPointRequirement' checks if the client has a certain number of quest points. Several quests require a minimum number of quest points in order to start the quest so this seemed like an easy one.

`QuestRequirement` checks if the client has completed a certain quest. As well, if the quest only requires the player to start the required quest but the client has finished that quest, this will return true.

`SkillRequirement` checks if the client has at least a required skill level. Some quests allow boosting of skill levels in order to complete certain parts. This requirement allows for that as well.

### DIFF
--- a/src/main/java/com/questhelper/requirements/CombatLevelRequirement.java
+++ b/src/main/java/com/questhelper/requirements/CombatLevelRequirement.java
@@ -3,6 +3,7 @@ package com.questhelper.requirements;
 import lombok.RequiredArgsConstructor;
 import net.runelite.api.Client;
 import net.runelite.api.Experience;
+import net.runelite.api.Player;
 import net.runelite.api.Skill;
 
 @RequiredArgsConstructor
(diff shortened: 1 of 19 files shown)


Zoinkwiz (changes requested):
This is really good! The cleanup of Operation is really nice. One point but otherwise happy to merge this in 😄.

Senmori:
Yes, a very good one.
That reason is I didn't know that was possible. I've replace it in the next commit.

Zoinkwiz (approved):

PR #137, opened 2021-01-13
Add General Requirements to all quests.

Senmori:
I need alcohol.

I went down the wiki's list of quests and their requirements and added the general requirements for each quest.
I might have missed some if they weren't there but at least the majority of them are done. The rest can be added as they are found.

https://oldschool.runescape.wiki/w/Quests/Requirements

Also,  one commit so the commit history isn't screwed up 👍


Zoinkwiz (approved):
Really good stuff, thanks!

PR #138, opened 2021-01-15
Implement more filters and ordering options for quests

Senmori:
Configuration: https://i.imgur.com/AmNSIi5.png

Result: https://i.imgur.com/ctjNX8T.png

***1. Filters***
There are 7 filters now.
They are all pretty self explanatory except one.

`BY_DIFFICULTY` works in conjunction with the `difficulty` configuration option to determine what difficulty of quest should be shown.

I have extracted the 'Show completed quests' option into a configuration option. This gives users more flexibility when determining which quests they want to see.
For example, if you wanted to see how many Novice quests you have done in relation to how many you have completed this is possible which this option.
It also allows people to see their progress through the Optimal Quest Guide easier.

****2. Ordering****
The only thing that changed with the ordering was the addition of `Z_TO_A` which is just a reversed `A_TO_Z`.
I also used Comparators to allow for them to be used in stream and on lists much easier.

****3. Misc****
I added a new color for quests that are in progress. Additionally I think it'd be nice to export the quest colors (green, white, in progress color) to the configuration.

Additionally, all quest filtering and ordering is now done exclusively in `QuestHelperPlugin`. There was a possibility using the new filters that you could try and call something on a quest that could only be done on the `ClientThread` which would result in an error.
`QuestSelectPanel`'s refresh method now has a `Map<QuestHelperQuest, QuestState>` field that allows us to pass in quest state's for all quests. This is what allows us to determine quest completion while off the `ClientThread`.


src/main/java/com/questhelper/QuestHelperConfig.java
```diff
@@ -38,14 +44,60 @@
...
+		OPTIMAL(QuestOrders.sortOptimalOrder()),
+		;
+
+		private final Comparator<QuestHelper> comparator;
+		QuestOrdering(Comparator<QuestHelper> comparator) {
+			this.comparator = comparator;
+		}
+
+		public List<QuestHelper> sort(Collection<QuestHelper> list) {
+			return list.stream().sorted(getComparator()).collect(Collectors.toList());
```
Senmori, line 61: You can call `sort` on collections directly, however that modifies that collection. This way allows us to keep that collection intact and get a new list in return.

src/main/java/com/questhelper/questhelpers/Quest.java
```diff
@@ -0,0 +1,59 @@
...
+ *  * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ *  * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ *  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ *  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+package com.questhelper.questhelpers;
+
+public interface Quest
```
Senmori, line 29: I didn't want to create a new class originally, however putting it in the `QuestHelperConfig` class made it very tedious to find anything.

src/main/java/com/questhelper/QuestHelperPlugin.java
```diff
@@ -310,21 +314,19 @@ public void updateQuestList()
...
+
+			List<QuestHelper> filteredQuests = allQuests
+				.stream()
+				.filter(q -> config.filterListBy().test(q, config))
+				.filter(q -> (config.showCompletedQuests() && q.isCompleted()) || !q.isCompleted())
+				.sorted(config.orderListBy().getComparator())
+				.collect(Collectors.toList());
+			Map<QuestHelperQuest, QuestState> completedQuests = quests.values()
+				.stream()
+				.collect(Collectors.toMap(QuestHelper::getQuest, q -> q.getQuest().getState(client)));
```
Zoinkwiz, line 327: This is a bit problematic for quests/miniquests with more complex definitions for whether they're started/completed (RFD sub-quests, if we add non-listed tasks such as Barb Training) as the checks don't extend to them (https://github.com/RuneStar/cs2-scripts/blob/e0bce7a2bc1e27a2910861a9e18140207587508b/scripts/%5Bproc%2Cquestlist_progress%5D.cs2). 

May be worth in QuestHelperPlugin having:

```
.collect(Collectors.toMap(QuestHelper::getQuest, q -> q.getQuest().getState(client)));
```

Instead be:

```
.collect(Collectors.toMap(QuestHelper::getQuest, q -> q.getState(client)));
```

Where `getState` by default calls getState of the quest as done above, but can be overriden. I believe the RFD sub-sections already have isComplete overriden in pretty much an identical way you'd want this to be, so should be easy to do. Would then also make this easily scalable in the future.

src/main/java/com/questhelper/QuestHelperConfig.java
```diff
@@ -112,11 +164,33 @@ default QuestOrdering orderListBy()
...
+		description = "Configures what to filter in the quest list",
 		position = 1
 	)
 	default QuestFilter filterListBy()
 	{
-		return QuestFilter.HIDE_DONE;
+		return QuestFilter.SHOW_ALL;
+	}
+
+	@ConfigItem(
```
Zoinkwiz, line 175: I'm not sure about having difficulty filtering as is currently implemented, it feels a bit confusing having to select two different filters in combination to get it working.

Looking at it now, I reckon having the filters separated to tick boxes would work better. This'd allow users to stack filters which could be nice as well.

I'd say sections would be 'Hide done', 'Hide quests', 'Hide miniquests', 'Hide can't do' 'Hide P2P', 'Hide F2P', then the current dropdown for difficulty filtering which has the additional 'All'.

WDYT?

Zoinkwiz (changes requested):
A couple points, would be good to hear your thoughts on them 😃

src/main/java/com/questhelper/panel/questorders/QuestOrders.java
```diff
@@ -202,4 +208,23 @@
 		QuestHelperQuest.ENCHANTED_KEY,
 		QuestHelperQuest.FAMILY_PEST
 	));
+
+	public static String normalizeQuestName(String questName) {
+		return QUEST_NAME_PATTERN.matcher(questName).replaceAll("").trim();
```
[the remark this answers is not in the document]
Senmori: The pattern only searches for those words at the beginning of a string. RFD doesn't have those strings. If you filter by RFD it does sort them

src/main/java/com/questhelper/QuestHelperConfig.java
```diff
@@ -38,14 +44,60 @@
...
+		/** Show all free-to-play quests */
+		FREE_TO_PLAY((q,c) -> q.getQuest().getQuestType() == Quest.Type.F2P),
+		/** Show all members' quests */
+		MEMBERS((q,c) -> q.getQuest().getQuestType() == Quest.Type.P2P),
+		/** Show all miniquets (all miniquests are members' only) */
+		MINIQUEST((q,c) -> q.getQuest().getQuestType() == Quest.Type.MINIQUEST),
+		/** Sort by difficulty */
+		BY_DIFFICULTY((q,c) -> q.getQuest().getDifficulty() == c.difficulty()),
+		/** RFD cause it ruins everything */
+		RFD((q,c) -> q.getQuest().getDifficulty() == Quest.Difficulty.RFD),
```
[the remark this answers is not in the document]
Senmori: I think we need it because it allows us to separate out RFD from the other quests. Due to how RFD quest state is stored as one varbit between what is basically 10 quests it's really annoying to try and separate what someone has completed from what they haven't. This way allows us to only show the 'Start - RFD' if they haven't even started RFD.
If someone has started RFD I'm not sure what it would show.

src/main/java/com/questhelper/QuestHelperPlugin.java
Senmori: When I get back home I'll change this

src/main/java/com/questhelper/QuestHelperConfig.java
Senmori: I'm actually not a fan of separating everything into checkboxes. 
Mainly because how often will you be stacking filters like that. I think the filter that are present are pretty much what's going to be searched for.
As well, I think it would just be too many options for the user to see at once. The only quests we really need to filter out are those that are completed (or not). The rest of the quests fall into one of the filter categories nicely.

I can add the 'All' difficulty option though, that's pretty easy.

src/main/java/com/questhelper/QuestHelperConfig.java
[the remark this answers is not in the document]
Senmori: Completed

Zoinkwiz (approved):
This is really nice, and works like a dream now 😄. I'll merge this in, then add in some logic for the RFD sub-quests being in progress.

PR #143, opened 2021-01-16
Cold War Fixes

julysfire:
- Added World Points to a lot of steps, this actually fixed one step that continuously crashed (enterAvalanche).
- Move talkToLarryAfterEmotes into the conditional step for returning to Relleka otherwise this would not progress.
- Moved "Yes" dialog step from talkToLarryOnIcebergWithSuit to tuxedoTime as the varbit progreses to this step before the dialog step.
- Removed readMissionReport and kgpBeforePingPong from the Panel as these steps are skipped and are not required for any progression within the quest.

As always, reach out if there are any questions/concerns/anything :)

Edit: Formatting on comments.


Zoinkwiz (approved):
Looks good, thanks!

PR #144, opened 2021-01-16
Add basic Quest Bank Tab solution

Zoinkwiz:
This adds a button to the bank which brings up an interface with the items you'll need for the currently active quest, with the items sorted into sections, and the number of items indicated.

Few things in terms of player-facing bits I'm unsure on:

1. The position of the button is far from ideal. I've put it at the top right to avoid it conflicting with any existing Runelite functionality, but it feels a bit rubbish. Any ideas on good placement + how it should interact are appreciated
2. Quantity overlay is okay, but can be bad for larger quantities, and over brighter items.
3.

### DIFF
--- a/src/main/java/com/questhelper/BankItems.java
+++ b/src/main/java/com/questhelper/BankItems.java
@@ -1,5 +1,5 @@
 /*
-	* Copyright (c) 2020, Zoinkwiz <https://github.com/Zoinkwiz>
+	* Copyright (c) 2021, Zoinkwiz <https://github.com/Zoinkwiz>
 
 	* All rights reserved.
 	*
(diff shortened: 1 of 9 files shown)


Senmori:
Just checked this out. Love the interface itself.
However, there are some issues I found.

1. NPE when switching from the questhelper tab to the 'View all Items' tab.
  - NPE: https://gist.github.com/Senmori/a9b56502e8c79789b7831436c4500fa4

2. There are some issues with large quantities interfering with the display as you've noticed:
  - https://i.imgur.com/dFYkAh4.png

As for icon placement, the only other place that would make sense to me where it would not interfere would be right next to the bank space icon.

PR #145, opened 2021-01-17
Add the ability to start creating developer tools

Senmori:
This is a simple change but it allows us to start talking about what tools we want to add for when users are working on QuestHelper.
Either when they are making a quest or just doing dev work in general.

I did a short test and confirmed this worked by adding to the overlay in the `QuestHelperOverlay` class.

I believe with this we can now start working on some useful tools for when quests need to be added/edited.

You could also a way for devs to 'step' through their existing steps without having to progress through a quest so they can make sure their zones/objects/npcs work as intended. Obviously these won't work for instances or places that only appear during the quest, but for most things this should remove the doubt that a particular step works.


Zoinkwiz (approved):
Looks good, thanks 😃

PR #146, opened 2021-01-17
Change how config filters are ordered.

Senmori:
Also added a config section to better define the different parts of the config.

The filters should now read like a sentence.
(e.g. Show all grandmaster quests in alphabetical order is equivalent to what's in the picture)

I was originally going to include completed quests right below the main filter drop down box but it ruins the way that section looks.

Picture: https://i.imgur.com/mbh2ZRn.png


Zoinkwiz (approved):
Looks really nice, the descending sizes of the sections is very satisfying.

PR #147, opened 2021-01-17
Add the basic framework for adding debug information

Senmori:
This will allow devs to have more information on screen while they are working on quests.

This PR implements the basic framework that will allow devs to add debug information on screen while working in dev-mode.
At the moment, all it does is display the current Var of the active quest, but devs can override that behavior in each quest using the correct methods.

See `QuestHelperRenderer` for which method to use.

`renderDebugOverlay` is for rendering overlay similar to how quest step overlays are rendered.
`renderDebugWorldOverlayHint` is for rendering overlays similar to how fishing spots/agility spots are rendered
`renderDebugWidgetOverlayHint` is for rendering overlay hints on widgets

### DIFF
--- a/src/main/java/com/questhelper/QuestHelperConfig.java
+++ b/src/main/java/com/questhelper/QuestHelperConfig.java
@@ -204,4 +204,15 @@ public interface QuestHelperConfig extends Config
 	{
 		return false;
 	}
+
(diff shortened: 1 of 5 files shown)


Zoinkwiz (approved):
Looks good, thanks

PR #150, opened 2021-01-19
QuestRequirement now displays custom display text.

Senmori:
For example, "Fairy ring access" looks better than "Fairytaile II - Cure a Queen In Progress"


Zoinkwiz (approved):
Looks good, thanks. Worth pointing out for the Fairy Ring access, I imagine a VarbitRequirement would be better, which specifies the exact value required for access for fairy ring access vs just completion/inprogress check of the quest.

PR #154, opened 2021-01-20
General code improvements

Senmori:
`QuestOverviewPanel` did not respect `NoItemRequirement`. `NoItemRequirement` now overrides `getColor` and will return the correct color on the panel.
See GIF: https://i.imgur.com/C0PBuLD.gif

`InventorySlots` is the new container for where whole inventories are checked.
`ItemSlots` now has a reference to `InventorySlots` for the three enums that require it.

`NoItemRequirement` now requires an `ItemSlots`. This lets us get rid of magic values in `NoItemRequirement`.
As well, we can compartmentalize the logic where it should be.

Moved the single slot inventory checks into `ItemSlots` and whole inventory checks into `InventorySlots`
Also, streams. :)

I replaced several for loops for stream where the logic itself was simple. Anything more complex than a boolean check, or incrementing a counter was not replaced.

`NoItemRequirements` was using the "wrong" super method. It was reporting a quantity of 1, so the side panel was showing 
`1 x Nothing in your <slot>` instead of `Nothing in your <slot>`

`ObjectStep` was cleaned up so it's easier to read the fill color.

### DIFF
--- a/src/main/java/com/questhelper/requirements/ItemSlots.java
+++ b/src/main/java/com/questhelper/requirements/ItemSlots.java
@@ -24,11 +24,16 @@
  */
 package com.questhelper.requirements;
 
+import com.questhelper.InventorySlots;
+import lombok.Getter;
+import net.runelite.api.Client;
+
+@Getter
 public enum ItemSlots
 {
-	ANY_EQUIPPED_AND_INVENTORY(-3, "inventory or equipped"),
-	ANY_INVENTORY(-2, "inventory slots"),
-	ANY_EQUIPPED(-1, "equipped slots"),
+	ANY_EQUIPPED_AND_INVENTORY(-3, "inventory or equipped", InventorySlots.EQUIPMENT_AND_INVENTORY_SLOTS),
+	ANY_INVENTORY(-2, "inventory slots", InventorySlots.INVENTORY_SLOTS),
+	ANY_EQUIPPED(-1, "equipped slots", InventorySlots.EQUIPMENT_SLOTS),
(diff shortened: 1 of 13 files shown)


src/main/java/com/questhelper/requirements/ItemSlots.java
```diff
@@ -43,21 +50,50 @@
...
+		}
+
+		int[] equipment = client.getLocalPlayer().getPlayerComposition().getEquipmentIds();
+
+		if (equipment == null)
+		{
+			return true;
+		}
+
+		return getSlotIdx() >= 0 && equipment[getSlotIdx()] <= 512;
```
Senmori, line 96: the `getSlotIdx() >= 0` check doesn't *technically* need to be there since the `getInventorySlots() != null` check is there, but better safe than sorry

src/main/java/com/questhelper/requirements/NoItemRequirement.java
```diff
@@ -24,106 +24,37 @@
...
-import net.runelite.api.ItemContainer;
 
 public class NoItemRequirement extends ItemRequirement
 {
-	public static final int ALL_EQUIPMENT_SLOTS = -1;
-	public static final int ALL_INVENTORY_SLOTS = -2;
-	public static final int ALL_EQUIPMENT_AND_INVENTORY_SLOTS = -3;
-	int slot;
+	private final ItemSlots slot;
+	private final int matchingItemID;
```
Senmori, line 34: I don't like magic values.

PR #162, opened 2021-01-24
Add filtering logic to LogicType and a couple other things.

Senmori:
Zoink, please remember to squash the commits. :)

`BankTabItem` now has a constructor that takes `ItemRequirement` to make it a bit nicer to instantiate it with an `ItemRequirement`.

Streamified several classes:
* `QuestHelperBankTagService`
* `FollowerItemRequirement`
* `FollowerRequirement`
* `ItemRequirements`
* `ItemRequirementCondition`

`PanelDetails` now doesn't discriminate against non-item requirements. 
`QuestStepPanel` now displays all requirements, even non-item ones.

Added javadocs to lots of Requirement classes.
* Created `EquippedItemRequirement` to check if a player has an item equipped in a specified slot

`ItemSlots` now actually checks for the presence of an item in a given slot, instead of just checking if the item in a slot was an item.

`Requirements` uses the new `LogicType` check as a first run to see if any unexpected scenarios cause it to no function as intended.
I tested the `Requirements` using all `LogicType` enum constants, but there are a lot of quests so I can't be sure I covered every scenario. This is why I did not implement the `LogicType` checks in `ItemRequirements`. There are many more `ItemRequirements` checks than `Requirements` checks.

`Spellbook` now has it's own check.
* Adjusted `SpellbookRequirement` accordingly.

And finally, `LogicType`.
`LogicType` now uses an internal `BiFunction` to apply a stream onto a predicate.
LEGEND:
- inputs -> stream elements
The implemented logic types are:
* `AND` - returns true only if all inputs match the supplied predicate
* `OR` - returns true if any inputs match the supplied predicate
* `NAND` - returns false if all inputs  match the supplied predicate. Otherwise returns true.
* `NOR` - returns true if all inputs do NOT match the given predicate
* `XOR` - return true if either, but not both, inputs match the supplied predicate. This only tests the first two elements of the stream.

I tried to think of a way to do XOR while checking the entire stream but that's pretty much impossible considering the logic gate.
You could split the requirements up into pairs and check it that way, but I believe at that point it would be better for the dev to handle that.


Zoinkwiz (approved):
Brilliant work as always :).

PR #165, opened 2021-01-25
Fix Knight's Sword not highlighting cupboard and step requirement

Senmori:
1. When you search the cupboard for the portrait, the cupboard wasn't highlighting when it was closed (different object id).
    * Also added new requirements `ZoneRequirement` and `NpcRequirement` to see if Sir Vyin is not in the room when you search it.

2. Some step requirements were wrong with thurgo after you give him the portrait. Remove the unnecessary portrait requirement and adjusted the display text to match.


Zoinkwiz (approved):
Nice, thanks :)

PR #167, opened 2021-01-26
Combine requirements and conditions.

Senmori:
1. Combined old `Requirement` and `ConditionForStep` by adding a new interface they both implement.
    * The new interface is called `Requirement`, and the old requirement class is now called `AbstractRequirement`. Name can be changed.

2. Moved quest conditions into the requirement package so everything is centralized.
3. Added javadocs to all new classes.
4. Added two builder classes for requirements.
    * `ComplexRequirementBuilder` makes building `ComplexRequirement` a bit more fluid.
    * `RequirementBuilder` is for building a non-standard `Requirement`. This should prevent us having to create new classes for simple requirements. 
5. Add `QuestUtil` for convenience methods for creating ArrayLists and LinkedList as well as an arraylist collector for streams.
6. Various code cleanup and improvements.


src/main/java/com/questhelper/panel/PanelDetails.java
```diff
@@ -57,7 +58,7 @@ public PanelDetails(String header)
 	public PanelDetails(String header, QuestStep... steps)
 	{
 		this.header = header;
-		this.steps = Arrays.asList(steps);
+		this.steps = QuestUtil.toArrayList(steps);
```
Senmori, line 61: `PanelDetails` has `#addSteps(QuestStep...)` so this has to be an arraylist.

src/main/java/com/questhelper/quests/recipefordisaster/RFDEvilDave.java
```diff
@@ -192,7 +193,7 @@ public void setupSteps()
 	{
 		List<PanelDetails> allSteps = new ArrayList<>();
 
-		List<QuestStep> steps = Arrays.asList(inspectEvilDave, enterBasement,
+		List<QuestStep> steps = QuestUtil.toArrayList(inspectEvilDave, enterBasement,
```
Senmori, line 196: Unfortunately, `Arrays.asList` doesn't return the `ArrayList` in `java.util.ArrayList`. It returns a fixed-size list so we can't add or remove anything from the list.

Zoinkwiz (approved):
Looks great! And thanks for fixing the ArrayList flops 👍

PR #183, opened 2021-02-03
Add live overlay text replacement and add QuestItem for ItemRequirement organization.

Senmori:
* `QuestStep` can now specify it's own overlay text. The method is not called `setPanelText` because it might be confused with the side panels.
* `Requirements` can specify a panel replacement if their `check(Client)` fails.
    * replacements can have their own replacements as well. Don't recurse your replacements!
* Added more logic to `ItemSlots` and `InventorySlots`.
    * Previously, they could only check if ALL the slots matched a predicate. Now they can check if ANY of the slots match.
    * `LogicType` can now compare `Number`. Internally, it uses doubles for the most precision but it's a trivial change.
* `DigStep` will now highlight the spade (if you have it) if the dig tile is in the client's scene.
    * As well, you can pass it an expected item id (or predicate) and if you do not have that item (or the predicate fails) it will keep trying to highlight the spade.
* `QuestRequirementPanel` tooltips can now support simple html.

`QuestItem` can now be used to specify alternate items for `ItemRequirement`s. This allows the overlay to display the item the user currently has instead of the generic requirement name like the side panel.

@Haavardaw That branch name. <3


src/main/java/com/questhelper/requirements/util/InventorySlots.java
```diff
@@ -71,4 +71,22 @@ public boolean checkInventory(Client client, Predicate<Item> predicate)
...
+	 * @return true if ANY of the items match the predicate
+	 */
+	public boolean contains(Client client, Predicate<Item> predicate)
+	{
+		return Arrays.stream(inventoryID)
+			.map(client::getItemContainer)
+			.filter(Objects::nonNull)
+			.map(ItemContainer::getItems)
+			.flatMap(Arrays::stream)
+			.anyMatch(predicate);
```
Senmori, line 90: This check uses `anyMatch` while the `checkInventory` check uses `allMatch`.

src/main/java/com/questhelper/requirements/util/LogicType.java
```diff
@@ -33,27 +33,36 @@
...
 	/** Returns true if all elements do not match the supplied predicate. */
-	NOR(Stream::noneMatch),
+	NOR(Stream::noneMatch, (n1, n2) -> n1.doubleValue() == 0.0D),
 	/** Returns true if either, but not both, inputs match the given predicate.
 	 * This only tests the first two elements of the stream.
 	 */
-	XOR((s,p) -> s.filter(e -> p.test(e)).limit(2).count() == 1),
+	XOR((s,p) -> s.filter(p).limit(2).count() == 1, (n1, n2) -> {
+		return (n1.doubleValue() > 0.0D && !(n2.doubleValue() > 0.0D)) || (n2.doubleValue() > 0.0D && !(n1.doubleValue() > 0.0D));
+	}),
```
Senmori, line 48: The `XOR` logic is weird. Normally you don't XOR when comparing but I did the best I could.
The logic flow is 
```java
if  (first_num > 0 && !(second_num > 0)) || ((second_num > 0 && !(first_num > 0))
```
There's probably a nicer way to do it, but that's the only way that makes sense for XOR'ing two numbers

src/main/java/com/questhelper/steps/DetailedQuestStep.java
```diff
@@ -392,49 +391,41 @@ public void onItemDespawned(ItemDespawned itemDespawned)
...
+				itemList.forEach(i -> {
+					requirements.stream()
+						.filter(Objects::nonNull)
+						.filter(req -> req.getClass() == ItemRequirement.class)
+						.map(ItemRequirement.class::cast)
+						.filter(ItemRequirement::isActualItem)
+						.filter(req -> req.getAllIds().contains(i.getId()))
+						.forEach(req -> tileHighlights.get(tile).add(i.getId()));
+				});
+			});
```
Senmori, line 425: So this does work. I opened 20-30 quests and received no errors.
If you have questions about it let me know. If you think it's too much then I can revert it.

src/main/java/com/questhelper/requirements/ItemRequirement.java
```diff
@@ -118,6 +123,15 @@ public void addAlternates(Integer... alternates)
...
 
+	public void addAlternateItem(int itemID, String text)
+	{
+		this.itemNameMap.put(itemID, text);
+		if (!alternates.contains(itemID))
+		{
+			alternates.add(itemID);
+		}
+	}
+
```
Senmori, line 134: Basically, if you directly add alternates you will use the default name provided via the constructor. If you add alternates via this method it will update the overlay accordingly.
The quantity of required item(s) stays the same.

src/main/java/com/questhelper/requirements/AbstractRequirement.java
```diff
@@ -45,6 +48,49 @@ public String getTooltip()
 	@Override
 	public void setTooltip(String tooltip)
 	{
-		this.tooltip = tooltip;
+		appendToTooltip(tooltip);
```
Senmori, line 51: Not completely sold on this. It makes sense for `setTooltip` to override what is already in the tooltip string.
Thoughts?

src/main/java/com/questhelper/requirements/ItemRequirement.java
```diff
@@ -60,6 +63,7 @@
 	protected boolean highlightInInventory;
 
 	protected final List<Integer> alternates = new ArrayList<>();
+	protected final Map<Integer, String> itemNameMap = new HashMap<>();
```
Zoinkwiz, line 66: I'm not really a fan of this way of doing it, it's quietly used just by specifying a quantity but nothing else for the initial name, which isn't ideal, and ends up with replication of ids, just 1 has more info.

I think what I'd prefer is we create a class, say 'QuestItem', which will define the ID/Name relationship, and replace alternates with that. 

We can then have a boolean which explicitely states to override the text, and uses `client.getItemDefinition(ITEM_ID).getName()` to get its name. WDYT?

src/main/java/com/questhelper/steps/DetailedQuestStep.java
```diff
@@ -82,9 +90,10 @@
 	protected List<WorldPoint> worldLinePoints;
 
 	@Setter
+	@Getter
 	protected List<Requirement> requirements = new ArrayList<>();
 
-	protected HashMap<Tile, List<Integer>> tileHighlights = new HashMap<>();
+	protected Multimap<Tile, Integer> tileHighlights = ArrayListMultimap.create();
```
Zoinkwiz, line 96: Didn't know this was a thing, nice 😄

src/main/java/com/questhelper/steps/DetailedQuestStep.java
Zoinkwiz: I think this one is getting a bit much, I think there's better ways for us to break it down and make it cleaner. Somehow the filters are about the same number of lines excluding brackets and the same depth 😅 .

src/main/java/com/questhelper/steps/DigStep.java
```diff
@@ -24,23 +24,62 @@
...
+		hasExpectedItem = InventorySlots.INVENTORY_SLOTS.contains(client, expectedItemPredicate);
+		if (!hasExpectedItem)
+		{
+			Player player = client.getLocalPlayer();
+			if (player == null) {
+				return;
+			}
+			WorldPoint targetLocation = worldPoint;
+			/*
+			 * Technically, since the grave is by the fence it would trigger if they came up to it from behind but
```
Zoinkwiz, line 77: Given this is a generic method, not just for one quest with a grave, I'd remove the comment.

src/main/java/com/questhelper/requirements/ItemRequirement.java
Senmori: Fixed. As discussed on discord.

Zoinkwiz (changes requested):
Couple points, and I've raised a few larger points in Discord.

src/main/java/com/questhelper/ItemCollections.java
```diff
@@ -534,7 +726,7 @@
 	// Logs
 
 	@Getter
-	private static final List<Integer> logsForFire = Arrays.asList(
+	private static final List<Integer> logsForFire = QuestUtil.toReversedLinkedList(
```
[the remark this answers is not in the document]
Senmori: Fixed.

src/main/java/com/questhelper/ItemDefinitions.java
```diff
@@ -0,0 +1,129 @@
...
+	@Getter
+	private final QuestItem nullItem = new QuestItem(-1, "I'm null. Fix me!");
+	private final Map<Integer, QuestItem> questItems = new ConcurrentHashMap<>();
+
+	protected void init()
+	{
+		//TODO: Add more commonly used items so they devs can save time creating those
+		// These items should be used in almost every quest
+		add(getNullItem());
+		add(new QuestItem(ItemID.COINS_995,"Coins"));
```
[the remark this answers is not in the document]
Senmori: Fixed. Technically?

src/main/java/com/questhelper/ItemCollections.java
```diff
@@ -81,45 +88,61 @@
...
+		ItemID.STARTER_BOW,
 		ItemID.OAK_SHORTBOW,
 		ItemID.OAK_LONGBOW,
+		ItemID.SIGNED_OAK_BOW,
 		ItemID.WILLOW_SHORTBOW,
 		ItemID.WILLOW_LONGBOW,
+		ItemID.WILLOW_COMP_BOW,
+		ItemID.MAPLE_SHORTBOW,
+		ItemID.MAPLE_LONGBOW,
+		ItemID.OGRE_BOW,
```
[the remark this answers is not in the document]
Senmori: Fixed. Reverted to original list. Will address in the future when we determine a new collection is needed.

Zoinkwiz (approved):

PR #187, opened 2021-02-05
Create SpellRequirement for easier spell manipulation.

Senmori:


src/main/java/com/questhelper/spells/Spell.java
```diff
@@ -0,0 +1,144 @@
...
+	{
+		this.groupId = 218;
+		this.widgetID = widgetID;
+		this.requiredMagicLevel = requiredMagicLevel;
+		this.operator = operator;
+	}
+
+	public Requirement getRequirements()
+	{
+		return operator.apply(StandardSpellBuilder.builder(this)).build();
```
Senmori, line 142: I left it as Requirement for now since I'll be doing more work on this. It will return the specific type of requirement that is decided upon.

src/main/java/com/questhelper/spells/StandardSpellBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+	{
+		requirementList.add(new ItemRequirement("", Arrays.asList(itemIDs), 1, equipped));
+		return this;
+	}
+
+	public ItemRequirements build()
+	{
+		String spellName = spell.name().toLowerCase(Locale.ROOT).replaceAll("_", " ");
+		String formattedName = WordUtils.capitalizeFully(spellName);
+		return new ItemRequirements(LogicType.AND, formattedName, requirementList);
```
Senmori, line 85: Eventually, I'd like to actually have a `MagicSpell` interface that all the spells implement so we can have their names, widget ids, icons, etc in one central place and then we can just adjust this builder to accommodate `MagicSpell` and that should enable us to build all spells with this builder.

src/main/java/com/questhelper/spells/StandardSpellBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+	{
+		this.spell = spell;
+	}
+
+	public static StandardSpellBuilder builder(Spell spell)
+	{
+		return new StandardSpellBuilder(spell);
+	}
+
+	public StandardSpellBuilder rune(int quantity, Rune rune)
```
Senmori, line 53: I decided on `(int, Rune)` because it reads easier. 
For example: 2 air runes, 2 water runes
That's a bit easier to read than:
air runes x 2, water runes x 2

This especially helps in `Rune` where it's very dense.

src/main/java/com/questhelper/spells/StandardSpellBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	public StandardSpellBuilder rune(Rune rune)
+	{
+		return rune(1, rune);
+	}
+
+	public StandardSpellBuilder item(int quantity, int itemID)
+	{
+		requirementList.add(new ItemRequirement("", itemID, quantity));
```
Senmori, line 66: I figured we could use `client.getItemDefinition` to look up the item name if this item has to be displayed.

src/main/java/com/questhelper/requirements/SpellRequirement.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
+import lombok.Getter;
+import net.runelite.api.Client;
+import net.runelite.api.Item;
+
+public class SpellRequirement extends ItemRequirement
```
Senmori, line 44: Reminder for me mainly.
TODO:
Implement requirement checking.
Create overlay text.
Highlight npc/objects.
Highlight widget icons.

Senmori (commented):
Remove forwarding method calls. Users can use `addSpell` on QuestStep.

Zoinkwiz:
Arrows for DetailedQuestStep aren't appearing.

Zoinkwiz:
Just did some testing again, it seems there's a few issues with recommended item not working correctly, not detecting if you have the needed items or not, and a few other things. The process I went through was:

I had a step which requires Wind Blast (1 death 3 air runes).
- At first it didn't recommend in the bank any air items, just the death rune.
- I took out some death runes and it updated to recommend an air staff, and adds it to the overlay item reqs
- I take out some air runes, but it doesn't register as me having the items needed for wind blast
- If I take out the air staff and drops it, it now still say I need an air staff, but the bank items is also showing the runes. Unfortunately taking out as many death and air runes at this stage still results in the 'Wind Blast' text being red

PR #189, opened 2021-02-06
Add Scorpion Catcher and recursive check for substeps

Abbotta4:
- Add Scorpion Catcher 
- Add `PanelDetails.contains` to check all substeps for the current step as well when choosing the panel to highlight.


Zoinkwiz (approved):
Looks good, thanks!

PR #197, opened 2021-02-08
Add Animal Magnetism

robertmaloney:
Notes:
- there is a dialogue path of buying one chicken at a time that isn't covered by this helper yet


Zoinkwiz (approved):
Looks great, thanks!

PR #206, opened 2021-02-10
Add the ability for NpcStep to show safespots for fights.

Senmori:
https://i.imgur.com/9OFfAe5.png

This adds the ability for `NpcStep` to add and show safespots during fights.
Multiple safespots can be rendered with each one getting a non-quest image related icon so it's clear what the box is for.


Zoinkwiz (approved):
Nice one, thanks!

PR #207, opened 2021-02-11
Arrow Toggle

nnapolitano:
Adds toggle in the settings pane for the flashing arrow on the minimap and overworld (enabled by default). This change does not affect the boxes around objects, but that could be added if desired. 

This feature is one i personally wanted, as I like the quest plugin, but found the arrows made quests too easy (and thus of a feeling of accomplishment). 

Tested through Misthalin Mystery, Could enable/ disable arrows for NPCs, objects, and objectives, and arrows would automatically stop & resume being rendered.


Zoinkwiz (approved):
Really like this, thanks for this!

PR #243, opened 2021-02-26
Biohazard- add 30gp to recommended items

nnapolitano:
While completing Biohazard I noticed that the helper states "Take the Plague Sample to the Chemist in Rimmington. You can take a boat from Ardougne Dock to Rimmington for 30gp."

30gp was not a recommended item, so this update just adds it


Zoinkwiz (approved):
Looks good, thanks :)

PR #273, opened 2021-04-01
added tai bwo wannai trio

longma1:
added tai bwo wanna trio

I'm sorry Zoinkwiz for making you review this, it probably will be as painful to review as it was to write.


Zoinkwiz:
Rebased + made a PR in the above PR.

PR #292, opened 2021-04-24
Fremmydiaries

Khatib81-GH:
easy tasks added for fremmy diaries


Zoinkwiz (changes requested):

src/main/java/com/questhelper/QuestVarbits.java
```diff
@@ -118,6 +118,10 @@
 	ACHIEVEMENT_DIARY_KARAMJA_MEDIUM(3598),
 	ACHIEVEMENT_DIARY_KARAMJA_HARD(3610),
 	ACHIEVEMENT_DIARY_KARAMJA_ELITE(4567),
+  	ACHIEVEMENT_DIARY_FREMENNIK_EASY(4507),
```
Zoinkwiz, line 121: These varbits don't appear to be right.

src/main/java/com/questhelper/QuestHelperQuest.java
```diff
@@ -224,6 +224,7 @@
 		Quest.Difficulty.ACHIEVEMENT_DIARY),
 	KARAMJA_ELITE("Elite Karamja Diary", QuestVarbits.ACHIEVEMENT_DIARY_KARAMJA_ELITE, 1, Quest.Type.ACHIEVEMENT_DIARY,
 		Quest.Difficulty.ACHIEVEMENT_DIARY),
+	FREMENNIK_EASY("Easy Fremennik Diary", QuestVarbits.ACHIEVEMENT_DIARY_FREMENNIK_EASY, 0, Quest.Type.ACHIEVEMENT_DIARY, Quest.Difficulty.ACHIEVEMENT_DIARY),
```
Zoinkwiz, line 227: A completion value of 0 means this won't run when it's not done, and will run when it is. Assuming it goes to 1 when complete, this should be 1.

Zoinkwiz (changes requested):
Currently the helper won't load, so I assume you've missed some definitions for requirements/steps.

PR #295, opened 2021-04-24
Update TheFeud.java

Kerpackie:
Some minor fixes.
Quest will no longer stop progressing passed the rug merchant.


Kerpackie:
Added quest requirement for Lost Tribe too :)

PR #300, opened 2021-04-28
Fix karamja medium diary quest requirement

Streammz:
In the Karamja Medium Diary, one of the requirements is intended to be the Shilo Village quest, but ended up being a duplicate line of The Grand Tree, probably due to a copy-paste error. This PR fixes these.


Zoinkwiz (approved):
Looks good, thanks!

PR #327, opened 2021-05-23
Add extra potato growing detail

Samywamy10:
Didn't know that I had to wait for the potatos to grow. Added some extra detail


Samywamy10:
@Zoinkwiz for review

Issue #334, opened 2021-06-05
Optimal Quest Order has changed with the addition of A Night at the Theatre


donicrosby:
The optimal quest order has been updated with the addition of A Night at the Theater. The optimal order in the plugin should reflect this change.

PR #338, opened 2021-06-06
Update Optimal Quest Order

Kerpackie:
Updates Optimal Quest list to updated order.

*Adds new quest, will need to be un-commented when quest is added.
*Adds all Achievement Diaries where required.
*Adds Stronghold of security for future addition.
*Adds Balloon Routes from post Enlightened Journey for future Generic Addition.


rypete:
This will resolve Issue #334

PR #377, opened 2021-06-23
Fix Amelia location in DS2

Kerpackie:
Fixes Amelias location in DS2, this should stop players wondering about, and getting lost. (Reported at least 2x daily)

Added Slayer Req for Fally Hard.


Kerpackie:
image

Zoinkwiz (approved):

PR #394, opened 2021-07-10
Achievement diary varrock

Obasill:
Varrock diary main functionality. Also added collections for high and low tier essence(for later use in diaries), coins didn't register large stacks so also created a collection, lastly added earth altar access collection


jcirillz (approved):

Zoinkwiz (approved):
Looks great, thanks 😄

PR #413, opened 2021-08-02
A Night at the Theatre

rypete:
A Night at the Theatre quest for the reworked parts. Includes everything except for ToB battle guides. The entire last part for ToB is entered under one step, "Complete Theatre of Blood"


Zoinkwiz (approved):
Looks great!

PR #423, opened 2021-08-11
Diary conversion to new panel system.

Obasill:
Conversion to new panel system for diaries. Also misc fixes to things I notice along the way.


Obasill:
Side note: why am I 45 commits ahead? The changes only happened in the latest commit.

Zoinkwiz (approved):
Looks good, thanks!

PR #424, opened 2021-08-11
Fally Diary fixes and Conversion

Obasill:
Converted fally diaries to new system. Added quest req where needed and made minor fixes


Obasill:
Added frem diary to this one. Should be all diaries except varrock changed. Did you want me to do that one or were you going to while you review it?

Zoinkwiz (approved):
Looks great, so nice seeing the new structure working! Thanks 😄

PR #430, opened 2021-08-23
Just some typos and a duplicate skills necklace requirement

Haavardaw:


rileyyy:
It might be beneficial to add a more descriptive title; something along the lines of "Rename Draken's Medallion to Drakan's Medallion" instead of "Just some typos".

For reference, the Skills Necklace change will also cause breaking changes as it is still being references on line ~818 as a recommended item.  It would also be encouraged to remove the declaration for it on line ~24 as that would have made the null reference a bit more apparent.  

To be honest though,  I think the duplicate skills necklace recommendation should be removed from the "Staff for Fairy rings or a Skills Necklace" as its a bit clearer for them to be listed separately.

PR #433, opened 2021-08-25
Missing Trowel requirement for the dig site

Haavardaw:
Just added Trowel to requirement list


rileyyy:
It is.  A trowel should be obtained after a player completes the[ Level 1 exam](https://oldschool.runescape.wiki/w/The_Dig_Site#Level_1).  It is the same time the player is awarded the Level 1 certificate

PR #435, opened 2021-08-29
Replace "Chaos Altar" step with "Chaos Temple"

deadmeu:
The name of the building containing the Wine of Zamorak is called the "[Chaos Temple](https://oldschool.runescape.wiki/w/Chaos_Temple_(Asgarnia))". Having the step refer to the "Chaos Altar" implies the [Chaos Altar](https://oldschool.runescape.wiki/w/Chaos_altar) in level 9 wilderness. This change brings the name in line with what is used on the OSRS Wiki for this [diary task](https://oldschool.runescape.wiki/w/Falador_Diary#Medium) (Falador medium - step 2).


rileyyy (approved):
Changes look good, though maybe it could be squashed into a single commit, since they are achieving very similar, concise tasks?

deadmeu:
> Changes look good, though maybe it could be squashed into a single commit, since they are achieving very similar, concise tasks?

I agree - I had missed some changes in the first commit. There should be an option to squash and merge at the same time, see https://github.blog/2016-04-01-squash-your-commits/

PR #443, opened 2021-09-03
Achievement diary lumb

Obasill:
Initial setup of Lumbridge diary, couple todos that I'll prob get back to when I'm done the rest of the diaries. Mostly just reusing collections from the varrock PR, made a really long and dumb gloves collection which was fun I guess. Nothing special in particular.


Obasill:
got merged but was remade by @Zoinkwiz because my branch had insane conflicts

PR #474, opened 2021-10-09
Talk to Lady of the Lake fix

Brok-Enwings:
Seems to be 5 and not 4 https://i.imgur.com/WQIV2x0.png


Zoinkwiz (approved):
Good spot, thanks for the fix!

PR #483, opened 2021-10-12
Add step for digging up mud within MountainDaughter quest for clarity

LoganHollins:
In the Mountain Daughter quest, you need to rub mud on the tall tree to be able to climb over the tree.  If you don't have any mud it isn't extremely obvious how to get the mud (Although it does have some text mentioning you need to dig it up in the item requirements, there isn't an explicit arrow / step pointing to where you need to dig), so I added a step that explicitly calls out the need to dig up mud if you don't have any and haven't rubbed the mud on the tree already. 

This is my first contribution to the plugin so if anything is wrong / should be changed / could be better, please call it out.


Zoinkwiz (approved):
Looks great, made an extra step to separate climbing + using mud as well. Not sure how I missed this varbit, thanks!

PR #484, opened 2021-10-12
Fixies

Obasill:
fixing lumb hard varpreqs, added rope to kand med (was previously on step), fixed quest order on lumb med, added additional step to frem med which checks if the special attack is on before highlighting the door to throw the axe at.

Thanks to @Jagerblue and @Leaf on discord for the feedback!


Zoinkwiz (approved):
Looks good, thanks :D

Issue #478, opened 2021-10-11
Wanted! item requirement bug


Nightfirecat:
The quest requires either 20 rune essence or 20 pure essence, but cannot be completed with a mixture of the two. (ex. 10 rune essence and 10 pure essence) The Mage of Zamorak only accepts a full 20 of either one.

The quest helper currently indicates you have the needed items if you have a combination reaching 20, which isn't correct.

PR #486, opened 2021-10-12
Resolving Wanted! essence issue #478

AidanDuffy:
Issue #478 explains that the guide states the user has all required items if they have some combination of rune/pure essence that are unnoted and amount to something >= 20. The quest needs all 20 essence to be the same (either all rune or all pure essence), so I split the logic into an OR, much like the commorb vs 10k requirement.


Zoinkwiz (approved):
Looks good, thanks!

PR #570, opened 2021-11-22
Achievement diary wildy

Obasill:
Wilderness diary  completed with rewards. Few things needed to be fixed / confirmed:

- Need someone with the completed Elite diary to give me their output of their "::getvarp 1193" to conclude the bit values of the elite tasks. 
- Resolve the issue of doing multiple bosses within one task, I'm a skiller so I'm unable to kill the bosses to gather the varb changes. 
- Check if another chat message requirement is necessary for the rune scimmy step in Elite. I don't have the necessary smithing lvl to complete this task.


Zoinkwiz (approved):
Looks great, thanks! :D

PR #576, opened 2021-11-25
Mournings End pt 1 & 2 changes

Obasill:
Finally got the ranged lvl to do these quests. Just some minor changes for clarity and direction.


Zoinkwiz (approved):
Looks good, thanks!

PR #587, opened 2021-12-01
Update General requirements in Fremennik Hard Diary

Felanbird:
edit: This PR now does the following:

- Gray's out the Spellbook requirement as being in 2 different spellbooks is not possible, and adding the quest requirement is redundant.
- Adds missing quests from `General Requirements` section
- Sets Giant Dwarf status from `COMPLETED` to `IN_PROGRESS` as per the wiki

It's not possible to be in 2 different spellbooks at the same time (not including Spellbook Swap), so the current check will always fail
(I should note this "fail" is only visual, it worked as intended when showing the diary when searching `Show Meet Reqs`)

~~I've taken the approach to checking for only `LUNAR_DIPLOMACY, QuestState.FINISHED`, as the Normal spellbook is always unlocked
Unsure if there's a better approach~~

Before
image

After
image

### DIFF
--- a/src/main/java/com/questhelper/achievementdiaries/fremennik/FremennikHard.java
+++ b/src/main/java/com/questhelper/achievementdiaries/fremennik/FremennikHard.java
@@ -249,8 +249,7 @@ public class FremennikHard extends ComplexStateQuestHelper
 		req.add(new SkillRequirement(Skill.THIEVING, 75, true));
 		req.add(new SkillRequirement(Skill.WOODCUTTING, 56, true));
 		req.add(new ComplexRequirement("Normal and Lunar spellbooks",
-			new SpellbookRequirement(Spellbook.NORMAL),
-			new SpellbookRequirement(Spellbook.LUNAR))
+			new QuestRequirement(QuestHelperQuest.LUNAR_DIPLOMACY, QuestState.FINISHED))
 		);
 		return req;
 	}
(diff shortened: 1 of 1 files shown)


Obasill:
If anything this should just be grey'd out. Since it's under general requirements and you're right obviously can't be on two spellbook at once.

Felanbird:
> If anything this should just be grey'd out. Since it's under general requirements and you're right obviously can't be on two spellbook at once. We already have the quest requirements so doing this doesn't add to the requirement list.

@Obasill Is switching this from a `ComplexRequirement` to an `ItemRequirement` reasonable or is there something better that I'm not thinking of. 

Was just going to switch the whole section to `req.add(new ItemRequirement("Normal and Lunar spellbooks", -1, -1));`

Obasill:
You're right that should be fine. It looks like I missed adding the quests to the general requirements. So toss these after the spellbook `req.add`

`req.add(eadgarsRuse);`
`req.add(lunarDiplomacy);`
`req.add(giantDwarf);`
`req.add(fremIsles);`
`req.add(throneOfMisc);`

Also giant dwarf needs to be in progress rather than finished.

Obasill:
Looks good, but it's giant dwarf that needs to be in-progress. From the wiki:
image

Zoinkwiz (approved):

PR #592, opened 2021-12-04
Add the option to order quests by release date

Mike-U5:
Adds the option to filter quests by release date


Zoinkwiz (approved):
Looks great, thanks!

Issue #599, opened 2021-12-09
Morytania hard diary gets bugged (after Salve shortcut step)


Scape070:
The plugin doesn't count completing the diary step "Use the shortcut to get to the bridge over the Salve" and gets stuck. Even if you open and close the diary menu, log in and out of game, or disable/enable the plugin, it still stays stuck there.

![2021-12-09_19-35-23](https://user-images.githubusercontent.com/95359977/145455889-a81441d4-59c6-4524-ad46-36c7c7c902e9.png)

Kerpackie:
Duplicate of #596 Currently working on resolving this!

Issue #596, opened 2021-12-07
Morytania Hard Dairy entry not flagging as completed


ZenithStar:
The entry "Use the shortcut to get to the bridge over the Salve" in the mort hard diary does not properly flag as completed, even upon opening the diary and client restart.

Issue #594, opened 2021-12-06
Easy wildy diary error


Avenandi:
The easy wildy diary lists that you need access to earth altar when it is in fact the need to access the chaos altar. the error exists in the item required tab.  And in the bring the following items tab in the chaos temple tab.

cheers! osy.

PR #600, opened 2021-12-09
Achievement Diary Minor Fixes

Kerpackie:
Fixes incorrect bit set in Mory Hard. Ref: #599 #596

(Thank you Haavard) 

Fixes the incorrect bit set for Salve Task in Mory hard.

Results based on Osu (Thank you Haavard):
`1180: 1111111111111111111111111111110` Value: `2147483646`
`1181: 111111110` Value: `510`

Fixes incorrect RC altar in Wildy Easy. Ref: #594


Kerpackie:
Ahh, dunno why its taking all those old commits - considering it was a new branch. I can resubmit or you can cherry pick the commits. Idm.

PR #614, opened 2021-12-16
Fix: optimal quest ordering

### LINKED ISSUE #609: Optimal Ordering - Desert Medium Diaries shows before Enakhra's Lament
The Camulet from Enakhra's Lament is required for Desert Medium Diaries, but the optimal ordering shows the diaries first:
image

image

binury:
Updates optimal order list to sync up with most recent version of [Optimal quest quide](https://oldschool.runescape.wiki/w/Optimal_quest_guide#Quests)

Fixes #609


src/main/java/com/questhelper/panel/questorders/QuestOrders.java
```diff
@@ -54,7 +54,6 @@
 		QuestHelperQuest.WATERFALL_QUEST,
 		QuestHelperQuest.TREE_GNOME_VILLAGE,
 		QuestHelperQuest.MONKS_FRIEND,
-		QuestHelperQuest.CLOCK_TOWER,
```
rileyyy, line 57: If I recall correctly, I was the one that moved this quest up in the list.  While it is not actually mandatory to be in a specific location, it is a short/novice quest which seems more at place near Monk's Friend.  My reasoning for this, aside from duration or difficulty, is that the 500 gp would be more useful to novice players at this point, and the quest point it gives would allow a player to ignore something else that they may not want to do before needing a required amount of qps.

binury: In that case, I think…

Option 1: _submit that suggestion to the wiki_, so that the plug-in's order will not diverge

Option 2: add code comment to notate the difference and reasoning

binury:
@Zoinkwiz need changes?

Zoinkwiz (approved):

PR #620, opened 2021-12-17
Another slice of ham corrected quest step.

Tolvinas:


Tolvinas:
seems like this is already fixed

PR #639, opened 2021-12-28
Update DesertTreasure.java to fix mistake in instructions.

SteveWick:
You do not need to make two trips to the bank like it says! Items can be brought noted.


jcarbelbide:
Confirmation from wiki that items can be noted:

Confirmation from wiki that items can be noted

[Wiki link](https://oldschool.runescape.wiki/w/Desert_Treasure/Quick_guide)

PR #656, opened 2022-01-03
Update Morytania Hard Diary Req's

Drakshad:
Watermelon seeds is not visible in the broader item req's, but is visible in the individual task. Just adding to broader item req's.


Zoinkwiz (approved):
Looks good, thanks.

PR #661, opened 2022-01-05
Add config to disable world line rendering

j-wreford:
The same should be added for other overlay hints, such as inventory item highlighting. Unfortunately I do not have an OSRS account in a position to test this, so just world lines it is.


Zoinkwiz (approved):
Looks good, thanks!

PR #673, opened 2022-01-14
More quest changes

Obasill:
handful of fixes to late game quests. Largely changed the path for MM2 sabotage, each step now has a route. Without any failures it is linear path to follow compared to the clusterfuck that was it earlier. Also added a 'reset' path +  step in the even someone does get caught they aren't completely lost on where to go. Sidenote: there are two 'styles' of route being used right now and a style someone else made for satchel 3+. The two styles are the movement to safe spot and solid line. It was a bit weird and somewhat confusing when I went through the platform. I'll leave that up to you if you want to change it.

Other changes are mostly just for clarity or fixing incorrect things.

### DIFF
--- a/src/main/java/com/questhelper/ItemCollections.java
+++ b/src/main/java/com/questhelper/ItemCollections.java
@@ -1474,11 +1474,11 @@ public class ItemCollections
 
 	@Getter
 	private  static final List<Integer> ardyCloaks = ImmutableList.of(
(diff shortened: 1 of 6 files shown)


Zoinkwiz (approved):
Looks good, thanks! Hopefully I can test out the new path with the release of leagues 😄

PR #691, opened 2022-01-26
Fix bridge requirements

Pyragon:
Each side of the bridge only requires 30 steel nails each.

Was confusing to me after repairing the first side, and the panel told me I required 60 steel nails for the second, even though it showed only 1 plank required. I ended up running to the bank and making another 30 steel nails before coming back.

This pull request makes that consistent with showing just 1 plank, and just 30 steel nails per side of the bridge.


Obasill:
You're better off just removing the quantity from the initialization and use .quantity(30) on each side.

Pyragon:
Will add another commit in the morning for that. 3am atm

Zoinkwiz (approved):

PR #694, opened 2022-01-28
More fixes for Zoink to go through :+1:

Haavardaw:
Quite a fix things for zoink to check

Not sure if the Ardy hard part is correct but I think it should be


Obasill:
In regards to the Ardy changes, they are good. One exception however:

`moveToPlatform = new NpcStep(this, NpcID.JEB, new WorldPoint(2719, 3305, 0),
			"Talk to Jeb to travel to the Fishing Platform.", smallFishingNet);`

Is correct but if the user hasn't completed the Slug Menace quest then they will see the travel option with Holgart. So just add him as an alternative NPC instead of what you have proposed.

Obasill (commented):
In regards to the Ardy changes, they are good. One exception however:

`moveToPlatform = new NpcStep(this, NpcID.JEB, new WorldPoint(2719, 3305, 0),
			"Talk to Jeb to travel to the Fishing Platform.", smallFishingNet);`

Is correct but if the user hasn't completed the Slug Menace quest then they will see the travel option with Holgart. So just add him as an alternative NPC instead of what you have proposed.

Obasill (approved):

Zoinkwiz (approved):

PR #698, opened 2022-01-31
Update LumbridgeHard.java

### LINKED ISSUE #696: Hard Lumbridge Diary - Natures missing
Nature runes are mentioned in the Bones to Peaches step but not in the general item requirements before starting. Should state that x2 Nature Runes are needed for the Lumbridge Hard Diary. Only other issue noticed was that Duel Arena was in lowercase for the step to use the altar with Smite on but that's being picky.

Felanbird:
- Capitalized Duel Arena & Daeyalt essence
- Adding missing `natureRune` in ItemRequirements
- renamed `ess` variable to `essence` & declared the requirement of `28` once
- declared the rune requirements for `waterRune` & `natureRune` once
- moved earthRune requirements to existing `earthRune4` & `earthRune10` variables
- declared `coins` variable once for barrows glove purchase, also moved it to the top of the ItemRequirements (looksout of place in the middle)

Fixes #696

also please squash and merge this, I seperated the commits very specifically so I could keep track of the changes, but it will look terrible in git history without squash&merge

### DIFF
--- a/src/main/java/com/questhelper/achievementdiaries/lumbridgeanddraynor/LumbridgeHard.java
+++ b/src/main/java/com/questhelper/achievementdiaries/lumbridgeanddraynor/LumbridgeHard.java
@@ -71,8 +71,8 @@ public class LumbridgeHard extends ComplexStateQuestHelper
 {
 	// Items required
 	ItemRequirement bones, earthRune4, earthRune10, earthRune14, waterRune, natureRune, fairyAccess, axe, goldBar,
-		cutDiamond, amuletMould, ballOfWool, cosmicRune, diamondAmuletU, diamondAmulet, miningHelm, tinderbox, coins, ess,
-		cosmicAccessOrAbyss, bellaSeed, seedDib, spade, rake, gloves, earthRune;
+		cutDiamond, amuletMould, ballOfWool, cosmicRune, diamondAmuletU, diamondAmulet, miningHelm, tinderbox, coins,
+		essence, cosmicAccessOrAbyss, bellaSeed, seedDib, spade, rake, gloves, earthRune;
 
 	// Items recommended
 	ItemRequirement ringOfDueling, gamesNeck, dorgSphere, lightsource;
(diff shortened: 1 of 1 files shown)


Obasill (approved):
Aesthetics aside, what purpose is there in changing variable names and replacing `.quantity(28)` into ` , 28)`? I don't mean to condescend, what I'm trying to say is that if it's not broken don't 'fix' it. 

Your time would be better utilized changing things that are incorrect / broken like the capitalization and nature rune changes you made. You've clearly a keen eye for that kind of stuff and I know I'm bad at that so I'm certain you'll be able to find more of my mistakes.

Felanbird:
I won't deny that I made the changes because I was already in the file for the mentioned issues.

I want a better consistency for this sort of thing, specifically for implementations by future contributors, as I borrowed a standard of code from another file for one of my first pull requests to this repo, and I believe declaring the values once, is a better consistancy.

Also for the same reasoning, I changed the variable name because it took me way too long to realize what an `ess` was.

Obasill:
If you want consistency that's great but you're going to have to go through every diary (except falador) and change those as well. Right now you're introducing a different way to code things in an otherwise consistent set of files, which makes it more confusing for future contributors. 

I don't know which file you're basing this standard on but wrote all the diaries (except karamja and fally) and I based my code off of how zoinkwiz did the karamja diary.

Felanbird:
While the diaries themselves might be fairly consistent, (although this exact file already has both variations [earth4 , 10 & 14]), many of the quests have the implementation I'm suggesting
`A Souls Bane, Biohazard, Bone Voyage, Cabin Fever, Cold War, Contact!, Grim Tails, Wanted!` from quick searching.

Although the standard may have been copied from the Karamja diary, it just doesn't seem like a good standard to uphold, declaring the value once instead of in every single mention of the variable just seems.... way easier and way more correct.

I split all the commits up like this to better keep track of them, but that way of committing is very useful for reverting the changes, so if the old standard is seen as better reverting this PR to just capitalization & missing variable can easily be done, but I do still think this is a better standard.

and if the other files need to be changed then so be it 😁

Obasill:
The reason I made multiple earthrune variables for this diary and defined the quantity in the initialization was to differentiate them for the two steps that used earth runes (power amulet and bones to peaches). That was because previously it would always show you needed 14 earth runes regardless of whether or not you've completed one of the two steps. 

I can't speak for zoinkwiz but the reason I use things like .quantity and .highlight is because it's a lot more dynamic. There are some tasks that have multiple steps that both require the same item but in different quantities. Hell, if you look at some of the old quests there are entirely new objects for things like coins / runes that need to be used multiple times or be highlighted. 

I've already approved the changes and don't mind them. It just seems like a lot of busy-work.

Zoinkwiz (approved):

Felanbird:
> I think .quantity can be nice in scenarios where you need varying quantities at different stages of the helper. 

Yeah I started playing around with moving more files to this, and whenever the quantities are varying, `.quantity()` just makes more sense. (seems to be very common with coins in Easy diaries.)

PR #704, opened 2022-02-03
Fix Legends Quest vanilla pod suggestion

Felanbird:
Fixes #701

PR #710, opened 2022-02-08
Fix for a typo in Legends Quest

duckdotapk:
Correcting a step to specify that the Vanilla Pods are in the south *west* of the Kharazi Jungle, as shown in the image below:

image
image

Also adding a full stop to the end of this step.


Felanbird:
#704

Zoinkwiz (approved):

PR #714, opened 2022-02-13
Fix Animal Magnetisim missed mistake

Felanbird:
I remember seeing my suggestion was acknowledged but I never re-looked at the code


Felanbird:
@Zoinkwiz 👁️

Zoinkwiz (approved):

PR #715, opened 2022-02-13
Update HorrorFromTheDeep.java

Felanbird:
As suggested in the original PR #691, the `.quantity()` approach is better for this, but I assume the PR author forgot to implement the changes after he agreed with this suggestion, so I'll just sneak them in.


Pyragon:
Shoot. My bad, it was like 3am and I 100% forgot by morning.

Thank you for fixing and pushing.

Zoinkwiz (approved):

PR #731, opened 2022-03-03
Achievement Diary Kourend

Obasill:
Fixed up the Kourend diary (original code by @rileyyy). Diary varps are correct took some black magic fuckery to figure out some of them. I don't know why but they aren't in bit order like the rest of the diaries. 

Everything is tested and fairly straight forward in terms of a diary.


Zoinkwiz (approved):
And that's the last diary completed! Amazing work, you've been so brilliant on getting these all done! Hopefully we can have them all live ASAP.

PR #732, opened 2022-03-04
shorten steps by not requiring rope and going into the barrel after talking to hudon

chestnut1693:
I was doing Waterfall Quest and i didn't bring a rope. I talked to Hudon and then clicked on the rock. I climbed out of the water and did something else. After a while I came back with rope to try once more. But before that, Quest Helper guided me to the house to read the book instead of guiding me to use rope on the rock and tree and to get in the barrel.

I haven't tested these changes to see if they have the desired effect.


chestnut1693:
It should be noted that there is fall damage involved doing it this way.

Zoinkwiz:
I believe you certainly used to take 10 damage or so, not sure if you still do. Would want to test that first before changing this as we don't want to get new accounts killed haha.

PR #739, opened 2022-03-17
Add more planks to Item Requirements

gare-bear:
Added Oak, Teak, and Mahogany planks as alternatives to satisfy item requirements.
Tested all planks in-game to make sure they were actually able to be used; checked both crossing into the island, and back to the mainland.


Zoinkwiz (approved):
Nice, thanks!

PR #749, opened 2022-04-01
"Snail" to "blamish" shell in easy Morytania diary

Frosty-J:
The snail shell is a drop in Temple Trekking. I may or may not have just done Temple Trekking until I got a snail shell, only to find it is not required and cannot be used for the task diary.


Zoinkwiz (approved):
Sorry for the misleading name but thanks for the fix!

PR #754, opened 2022-04-09
Add woodcutting requirement for ironman

saltz:
Add the 56 woodcutting requirement (splitting artic logs) for ironman players.


Zoinkwiz:
Thanks, added it in with an additional IronmanRequirement type I've just created so it won't incorrectly display non-ironman as not having the requirements.

PR #756, opened 2022-04-12
remove false pisc favor in easy diary requirement

buttwize:
You don't need any Piscarilius favor for the easy diary requirement. I assume it was added because of The Warrens step, but no favor is needed.
image


Zoinkwiz (approved):
Good to know, thanks!

Issue #753, opened 2022-04-07
Mourning's End Part 2


pwatts6060:
Catalytic Talisman can now be used instead of death talisman. It's a reward from Guardian's of the Rift and is tradeable. Item ID: 26798
The Wiki says the Catalytic tiara does not work for this step.

ekohvakka:
Another comment regarding Mourning's End Part 2: Add the teleport crystal to the list of recommended items for the quest.

Zoinkwiz:
Merged fix in, thanks!

Rerbun:
When doing this quest yesterday it still showed up red on the death altar puzzle with the Catalytic talisman in inventory

PR #760, opened 2022-04-15
Update Mourning's End Part II. Implement changes from #753 .

JoshuaPeddle:
Add Teleport Crystal as a recommended item. 
Add Catalytic Talisman as alternate for Death Talisman.


Zoinkwiz (approved):
Looks good, thanks!

PR #763, opened 2022-04-18
update quest point total to 287 

jacoblbeck:
Updated the required quest point total to 287


Zoinkwiz:
I've pushed another fix as of the new quest that came out today, so closing this.

PR #767, opened 2022-04-22
Couple fixes

Obasill:
Noticed a problem with overlapping dropdown on the panel from #755. The problem stems from the padding at the end of each option in the dropdown, so there might be a more suitable fix to this.

Before:
image

After:
image

Also a bunch of diary skill req adjustments. Previously they _sometimes_ included skill requirements for quests. Which didn't really make sense because in our quest guides they don't require the skills from leading quests.


Zoinkwiz (approved):
Looks good, thanks! 😄

PR #776, opened 2022-05-01
Additional details for ardougne med diary

JoshPaulie:
♥


Obasill:
You don't need to use the taverly hot air balloon, also I just noticed i mispelled 'balloon' in there

JoshPaulie:
@Obasill

Oh, okay. Just clarifying:

- The player can take **any** balloon to castle wars to complete to unlock the Adry Med Diary

- The grapple task can be completed by either climbing from inside yanile to outside, or from outside yanile to inside, BUT NOT up one side and down the same.

Obasill:
> * The player can take **any** balloon to castle wars to complete to unlock the Adry Med Diary
> * The grapple task can be completed by either climbing from inside yanile to outside, or from outside yanile to inside, BUT NOT up one side and down the same.

Correct

Zoinkwiz (approved):

PR #784, opened 2022-05-05
Add WidgetPresenceRequirement

claudiodekker:
Yesterday, @LlemonDuck mentioned to me that I can conditionally check whether a Widget is there or not (`null` vs. `Widget` instance) to detect eg. puzzle 'interfaces' being open.

After looking into it this morning I also noticed that there's already a `WidgetModelRequirement`, which does exactly that _except_ that it also checks for a Model ID to be matching, which for the sake of detecting a widget's presence/visibility is likely not needed, so it made me wonder whether it would make sense to add a `WidgetPresenceRequirement`  and have the `WidgetModelRequirement` become a subclass of that.

While this obviously doesn't add groundbreaking new logic, it does provide an easy-to-find Requirement for people new to the project, which does exactly what they would expect it to do, so I figured it'd be a good QoL change to submit.


Zoinkwiz (approved):
LGTM, thanks for adding!

PR #804, opened 2022-05-29
Small fix for step text

Fiffers:
Previously the text would have the user use the raw chompy on the NPC, but this would not progress the plugin to the next step correctly. Simply talking to the NPC works great, so I've changed the relevant step to instruct the user to do just that.


Fiffers:
Moved on to another quest and noticed another change to make the guide better, so I added a tooltip for a required item.

Zoinkwiz (approved):

PR #822, opened 2022-06-20
shark fishing spot is 1520

post-async:
diary was pointing to the lobby tuna swordfish spot, this is the shark spot.


claudiodekker:
Probably a good idea to also reference / use the Runelite built-in reference as alternative NPC ID's as to take advantage of upstream changes: https://github.com/runelite/runelite/blob/bed2220a426cbe77be3a39daeac3a98523260fe9/runelite-client/src/main/java/net/runelite/client/game/FishingSpot.java#L51

Zoinkwiz (approved):

PR #833, opened 2022-06-27
Add Quest: Temple of the Eye

Fincap:
This change would add the 'Temple of the Eye' quest to the plugin.

While there is still some polish work necessary, this quest helper should be functional for a typical player's quest playthrough, and accounts for most edge cases.

Notes that there are no quest helper steps for the Guardians of the Rift tutorial step except for directing the player back to the Temple if they leave/log out. This is because there is an in-game tutorial that should handhold the player well-enough.

Additionally, includes change to run configuration to prevent test client from loading external hub plugins. Credit to @LlemonDuck


Fincap:
My testing and refinement is complete. I understand I may not have caught everything, and may have broken some extra things with my additional changes, but it should be good to go.

Obviously open to feedback, happy to make any changes to anything that is clearly broken/inefficient/violates convention 😃

Zoinkwiz (approved):
Looks great, thanks!

PR #839, opened 2022-06-29
Karamja Medium Diary: add small fishing net as alternate item to raw karambwanji

dannil:
The `Raw karambwanji, or a small fishing net to catch some` text in item requirements were not correctly highlighted as green if you had a small fishing net in your inventory, as per the text instruction.


dannil:
@Zoinkwiz is this PR ok or should I change anything?

Zoinkwiz (approved):
Looks good, I think as a separate change this step should include fishing for karambwanji, and only use the raw karambwanji as an option for the catchKarambwan step.

PR #845, opened 2022-07-01
Add quest: Land of the Goblins

scatter-dev:
Added quest Land of the Goblins. I tested as I was playing but haven't been able to do a full runthrough after finishing the implementation.

KNOWN ISSUES: 
- Didn't catch the exact coords in the Yu'biusk instance at the end and you can't return to the instance after the quest, so I'll need someone's help to fix that. Might be "good enough" as is, but I just estimated.

OTHER CHANGES:
- There's a point in this quest where you have to accept your goblin name, and the dialog box isn't the same every time. I added regex support to WidgetChoiceStep to account for this.
- Added GREATER and LESS operations to Operation.java for use in conditionals, varbit reqs, etc.


Zoinkwiz (approved):
Looks great! Thanks 😄

PR #848, opened 2022-07-02
Update Prince Ali Rescue to align with new quest version

Fincap:
Prince Ali Rescue had some of its dialogue rewritten in the 27/04/2022 game update. This PR aligns the quest dialogue steps with the current version.

### DIFF
--- a/src/main/java/com/questhelper/quests/princealirescue/PrinceAliRescue.java
+++ b/src/main/java/com/questhelper/quests/princealirescue/PrinceAliRescue.java
@@ -30,6 +30,7 @@ import com.questhelper.QuestHelperQuest;
 import com.questhelper.Zone;
 import com.questhelper.panel.PanelDetails;
 import com.questhelper.questhelpers.BasicQuestHelper;
+import com.questhelper.requirements.RuneliteRequirement;
 import com.questhelper.requirements.item.ItemRequirement;
 import com.questhelper.requirements.Requirement;
 import com.questhelper.requirements.ZoneRequirement;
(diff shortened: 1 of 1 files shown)


src/main/java/com/questhelper/quests/princealirescue/PrinceAliRescue.java
```diff
@@ -169,10 +169,10 @@ public void setupConditions()
 	{
 		inCell = new ZoneRequirement(cell);
 		hasWigPasteAndKey = new Conditions(dyedWig.alsoCheckBank(questBank), paste.alsoCheckBank(questBank), key.alsoCheckBank(questBank));
-		givenKeyMould = new Conditions(true, LogicType.OR,
+		givenKeyMould = new Conditions(true, LogicType.OR,	// TODO quest journal widget text outdated
```
Fincap, line 172: The first two WidgetTextRequirements will not refresh the quest progress when opening up the Quest Journal, because the text has changed. I was unable to find the current version of the text so these conditions will never trigger until it is fixed.

src/main/java/com/questhelper/quests/princealirescue/PrinceAliRescue.java
```diff
@@ -169,10 +169,10 @@ public void setupConditions()
...
 			new WidgetTextRequirement(119, 3, true, "I have duplicated a key, I need to get it from"),
 			new WidgetTextRequirement(119, 3, true, "I got a duplicated cell door key"),
-			new WidgetTextRequirement(WidgetInfo.DIALOG_NPC_TEXT, "Pick the key up from Leela."),
+			new WidgetTextRequirement(WidgetInfo.DIALOG_NPC_TEXT, "I'll send it to Leela once it's ready."),
```
Fincap, line 175: This is untested, but should trigger the condition if the player does not close out of the NPC dialogue before this message is displayed.

Zoinkwiz (approved):
Looks good, thanks!

PR #889, opened 2022-07-25
Feature: Item -> Go to Wiki..

SpencerSKane:
Feature allows the user to right-click a Required or Recommended item in the panel opening a popup window. This popup window holds a link that, when clicked, sends the user to that items wiki page.

An items ItemID is assigned in the constructor and this feature grabs that id and goes to that ids wiki page. This means that all items(constructed with an id that is not -1) will have a corresponding wikipage and have the right-click option available without any further work.

-ItemRequirements with an assigned ItemID will be directed by that ItemID, unless a suffix has been assigned to it. 
-Item Requirements without an assigned ItemID (id: -1) will do nothing when right clicked, unless a suffix has been assigned to it.

`setUrlSuffix(String suffix)`
This method allows you to assign a suffix to redirect to a wikipage different than what the ItemID would send it to.
An example of this is included in the "Item wiki redirect" commit where I use this method on items that contain ItemCollections in the quest _Olaf's Quest._ 

**Example:** 
`axe = "Any axe", ItemCollections.getAxes() ` right clicking then clicking `'go to wiki..' `would send you to the crystal axe page. 
`axe.setUrlSuffix("Axe")` , now the action will send the user to the Axes page which lists all of the axes they could use which is more correct.


SpencerSKane:
Feature allows the user to right-click a Required or Recommended item in the panel opening a popup window. This popup window holds a link that, when clicked, sends the user to that items wiki page.

An items ItemID is assigned in the constructor and this feature grabs that id and goes to that ids wiki page. This means that all items(constructed with an id that is not -1) will have a corresponding wikipage and have the right-click option available without any further work.

-ItemRequirements with an assigned ItemID will be directed by that ItemID, unless a suffix has been assigned to it. 
-Item Requirements without an assigned ItemID (id: -1) will do nothing when right clicked, unless a suffix has been assigned to it.

`setUrlSuffix(String suffix)`
This method allows you to assign a suffix to redirect to a wikipage different than what the ItemID would send it to.
An example of this is included in the "Item wiki redirect" commit where I use this method on items that contain ItemCollections in the quest _Olaf's Quest._ 

**Example:** 
`axe = "Any axe", ItemCollections.getAxes() ` right clicking then clicking `'go to wiki..' `would send you to the crystal axe page. 
`axe.setUrlSuffix("Axe")` , now the action will send the user to the Axes page which lists all of the axes they could use which is more correct.

scatter-dev:
I'm wondering if it's worth the trouble of the `urlSuffix` part, for a couple reasons:

- To make use of it, someone would have to go through all the quests and look through all the requirements that aren't just a simple item ID, which is an unreasonable amount of work
- If it isn't set manually, then you get situations like "Any axe" taking the user to the wiki page for Crystal axe, which may lead people into thinking they need one (idk how likely that is really)
- Is it really that useful to direct the user to wiki pages like "Food" or "Axe"?

I think the ideal situation is general requirements like "food" or "any axe" don't show a wiki link at all. However, requirements like "amulet of glory" are just as general as far as the code is concerned because you could have any of uncharged/1 through 6. We could opt not to show the wiki link in all cases where more than 1 item ID is acceptable, but I can't decide on whether I think that tradeoff is worth it. Thoughts?

SpencerSKane:
I do see what you mean, I think it would be more correct to default to no link for item collections.

scatter-dev:
Having sat on it for a couple hours I think your implementation is best as is, and my concerns are not really worth giving up functionality for.

Zoinkwiz (approved):
Looks great! Thanks for the work on this :D

PR #891, opened 2022-07-26
GreatBrain - added tooltip to hammer

SpencerSKane:
Added tool-tip to hammer stating that Imcando Hammer should not be used as the hammer is given to Dr later in the quest. This change was requested on the Discord.


Obasill:
I think you'd be better off just changing the collection to ItemID.HAMMER instead of adding a tooltip. This makes it so there won't ever be people giving their Imcando hammer.

SpencerSKane:
You are correct, I completely overlooked that. Thank you!

Zoinkwiz (approved):

PR #936, opened 2022-09-30
Hide overlay on bank opened option

Obasill:
Adds option to hide quest step overlay when bank is currently open.


src/main/java/com/questhelper/QuestHelperConfig.java
```diff
@@ -200,6 +200,16 @@ default boolean showOverlay()
 		return true;
 	}
 
+	@ConfigItem(
+		keyName = "hideOverlayPanelBank",
+		name = "Hide overlay when bank opened",
+		description = "Chose whether the overlay should be displayed when bank opened"
```
tivervac, line 206:
```suggestion
		description = "Choose whether the overlay should be displayed when a bank tab is open"
```

src/main/java/com/questhelper/overlays/QuestHelperOverlay.java
```diff
@@ -33,25 +33,39 @@
...
 		this.plugin = plugin;
+		setLayer(OverlayLayer.ALWAYS_ON_TOP);
 		setPriority(OverlayPriority.HIGHEST);
 	}
 
 	@Override
 	public Dimension render(Graphics2D graphics)
 	{
+		Widget itemContainer = client.getWidget(WidgetInfo.BANK_ITEM_CONTAINER);
+		if (plugin.getConfig().showOverlay() && plugin.getConfig().hideOverlayBank() && !(itemContainer == null))
```
tivervac, line 65:
```suggestion
		if (plugin.getConfig().showOverlay() && plugin.getConfig().hideOverlayBank() && itemContainer != null)
```

Obasill: Is there a reason for this change? I'm not opposed to it, I'm simply curious.

tivervac: The semantics are the same, but it’s just a tad more readable

Obasill:
Updated branch so it's ready for merge

Zoinkwiz:
The overlay should go behind the bank interface now, so this shouldn't be necessary now.

behindinterface

Obasill:
Makes sense, which commit changed this? Cause I remember looking into this and didn't bother cause I found was a lot more work than this solution.

PR #966, opened 2022-11-15
Add tooltip for purchasing feather nearby

james1997hughes:
Spent 10 minutes punching seagulls to find out I misread their drop table. Added note for where to buy a feather in port sarim.


Felanbird:
It might be better if this was handled with a tool-tip, similar to `TheGrandTree`
https://github.com/Zoinkwiz/quest-helper/blob/3c97e0a29fc161cc1baa65fbf987e74888af851d/src/main/java/com/questhelper/achievementdiaries/westernprovinces/WesternMedium.java#L190-L195

`feather.setTooltip("Can be purchased at Gerrant's Fishy Business");`

james1997hughes:
Thanks for the review mate. Reverted previous string change & added tooltip as you described.

Zoinkwiz (approved):

PR #977, opened 2022-11-28
Fixes

Obasill:
Handful of changes I've noticed on the discord / issues list. I don't have admin on here or the discord so you'll have to clear the issue lists.

I can't do the recruitment drive / misc fixes because I don't have an account ready for those. If anyone has one let me know and I can take care of it. Otherwise i'll just wait until the next leagues which is hopefully soon™


Zoinkwiz (approved):
Looks really good, thanks as always for the great fixes!

PR #978, opened 2022-11-28
Add oak trees until lvl 35

jlereback:
Changed from normal trees to oak trees between 15 and 35 Woodcutting


Zoinkwiz (approved):
Looks good, thanks!

PR #984, opened 2022-11-30
Add Agility and update Woodcutting

jLereback:
Added Agility to Skill Helper and made Woodcutting usable for F2P


Zoinkwiz (approved):
Looks great, thanks!

PR #1033, opened 2023-01-13
Add highlight styles

mtesseracttech:
This adds new highlighting styles. 

Also cleaned things up a little bit here and there. Also seems the automatic formatting did not agree with how things were formatted in the files I worked on.

If there are other things that can reliably be highlit in other ways, I don't mind adding that as well.

### DIFF
--- a/src/main/java/com/questhelper/QuestHelperConfig.java
+++ b/src/main/java/com/questhelper/QuestHelperConfig.java
@@ -39,6 +39,7 @@ import net.runelite.client.config.Config;
 import net.runelite.client.config.ConfigGroup;
 import net.runelite.client.config.ConfigItem;
 import net.runelite.client.config.ConfigSection;
+import net.runelite.client.config.Range;
 import net.runelite.client.util.Text;
 
 @ConfigGroup("questhelper")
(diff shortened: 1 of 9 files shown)


src/main/java/com/questhelper/QuestHelperConfig.java
```diff
@@ -309,14 +358,81 @@ default boolean showTextHighlight()
...
+	)
+	default int outlineThickness()
+	{
+		return 4;
+	}
+
+	@ConfigItem(
+		keyName = "outlineFeathering",
+		name = "Outline feathering",
+		description = "Choose the thickness of target model outlines",
```
tivervac, line 428: Same description as Outline thickness?

src/main/java/com/questhelper/steps/DetailedQuestStep.java
```diff
@@ -122,10 +125,12 @@ public class DetailedQuestStep extends QuestStep
 	public boolean considerBankForItemHighlight;
 	public int iconToUseForNeededItems = -1;
 
+
 	public DetailedQuestStep(QuestHelper questHelper, String text, Requirement... requirements)
 	{
 		super(questHelper, text);
 		this.requirements.addAll(Arrays.asList(requirements));
+
```
tivervac, line 133: Not sure why you're adding new lines at the end of methods, is this part of the formatting rules? Doesn't look like it on first sight https://github.com/runelite/runelite/wiki/Code-Conventions

Zoinkwiz (approved):
Looks really nice! Thanks so much for this 😄

PR #1044, opened 2023-01-20
Add Mining and improved Agility help

jLereback:
Added guide for each obstacle instead of the first one in Agility Skill Training. Rellekka and Ardougne are left to do but I need higher Agility level to complete everything.

Added mining for free-to-play


src/main/java/com/questhelper/skills/mining/Mining.java
```diff
@@ -0,0 +1,191 @@
...
+			steelPickaxe, blackPickaxe, mithrilPickaxe, adamantPickaxe, runePickaxe);
+		ironStep.addTileMarker(IRON_POINT, SpriteID.SKILL_MINING);
+
+
+		mineCopperOrTin = new ObjectStep(this, ObjectID.ROCKS_11360, COPPER_POINT,
+			"Mine Copper- and Tin ore at South-east Varrock mine until 15 Mining. You can choose to drop " +
+				"the ores as you go or bank them in the eastern Varrock bank.", true,
+			ironPickaxe, steelPickaxe, blackPickaxe
+		);
+		mineCopperOrTin.addAlternateObjects(ObjectID.ROCKS_11360, ObjectID.ROCKS_11161, ObjectID.ROCKS_10943);
```
Zoinkwiz, line 157: Could we amalgamate this step with the 'copperStep' step, and add the Tile Icon to this step? Would mean people can always see where to stand and helps avoid needing to check where the player is in this helper.

Sames goes for iron and the members guide.

src/main/java/com/questhelper/skills/agility/Agility.java
```diff
@@ -44,58 +44,86 @@
...
 
-	QuestStep gnomeStronghold, draynorVillage, alKharid, varrock, canifis, falador, seersVillage, pollnivneach, rellekka, ardougne;
+	AgilityCourse gnomeStronghold, draynorVillage, alKharid, varrock, canifis, falador, seersVillage, pollnivneach;
+	ConditionalStep gnomeStep, draynorStep, alKharidStep, varrockStep, canifisStep, faladorStep, seersStep, pollnivneachStep;
+	QuestStep rellekka, ardougne;
 
 	@Override
 	public QuestStep loadStep()
 	{
+		setupCourses();
```
Zoinkwiz, line 71: With the All Items Helper, it runs 'setupRequirements' for each helper, and needs to have everything instantiated within it. To ensure the courses are setup as they're used for the reqs, moving this setup into setupRequirement should work fine!

src/main/java/com/questhelper/skills/agility/AlKharid.java
```diff
@@ -0,0 +1,133 @@
+/*
```
Zoinkwiz, line 1: I like the idea of having the QH arrow on the agility shortcut elements, but I'm not sure about having each step in the sidebar. What would you think of having in the sidebar just a single big step of 'Do the agility course in Al Kharid, starting on the north wall of the tannery' or w/e? This can be done with a parent ConditionalStep with the text, which is what's passed back to be displayed.

The individual step text could still appear on the overlay maybe, with a bit of a cleanup of the text to be more descriptive rather than just the use text of the obstacle course. Like 'Climb across the tightrope to the east.' and so on.

Zoinkwiz (changes requested):
Love seeing more skill helpers, thanks! A couple points for thought and discussion

src/main/java/com/questhelper/skills/mining/Mining.java
jLereback: I'm not exactly sure what you want me to do 🙈

Zoinkwiz: So, rather than having 2 steps, 1 being 'Stand next to copper' and then another of 'mine copper', could we have these steps combined into 1 step of 'Go to X location and mine copper and tin', with a tile marker added to the step to indicate where to stand?

PR #1048, opened 2023-01-23
Add Emir's PvP Arena Slayer helmet (i) variants

lezed1:
Add Emir's PvP Arena Slayer helmet (i) variants. I also added some missing existing variants for one collection.

I tested this on an account with a slayer helmet imbued via the PvP Arena. This does not work on the currently released version of Quest Helper, but it works on this new version!


Zoinkwiz (approved):
Looks good, thanks!

PR #1049, opened 2023-01-23
Deduplicate item collections

lezed1:


lezed1:
I combined this into #1048

PR #1052, opened 2023-01-29
Added lines to guide through maze in Waterfall Quest

quietdreamr:
Waterfall quest didn't have a guide to get through the maze, whilst Tree Gnome Village did. I took inspiration from the RatCatchers quest, which had a similar structure. Please let me know if there is anything that needs changing.


Zoinkwiz (approved):
Looks good, thanks!

Obasill:
Throwing error on line 265 / 267 because enterGnomeDungeonNoPath isn't defined or used.

quietdreamr:
> 

Ah, good catch. I didn't have a proper Java linter active whilst writing it.

This can be resolved by declaring `enterGnomeDungeonNoPath` as a `QuestStep` on line 76

https://github.com/Zoinkwiz/quest-helper/blob/master/src/main/java/com/questhelper/quests/waterfallquest/WaterfallQuest.java#L76

Obasill:
Or just removing it since it isn't being used, but yeah no problem. Just ran into it when I got up-to-date

PR #1056, opened 2023-02-03
Fixed compilation error in WaterfallQuest script

Syntribos:
Latest PR added a new variable and renamed the old one, but never declared the renamed var.

The step is now unused, so this PR removes the old variable to fix the build.


Zoinkwiz (approved):
Thanks for this :)

PR #1067, opened 2023-02-09
Adding Quest: Secrets of the North

Obasill:
My first quest, kinda weird since I've been contributing for over a year. I didn't have any issues beyond handling most of the quest is in an instance, so if the player leaves the instance I wasn't able to manage to direct the player back into the instance.


Zoinkwiz (approved):
Made some changes, but overall looks good thanks!

Issue #1080, opened 2023-02-18
Secrets of the North - Quest guide hyperlink is wrong


zahti:
The quest guide hyperlink is showing Dragon Slayer II - likely just a copy/paste error:
![image](https://user-images.githubusercontent.com/38710634/219821881-4804a5f0-171d-4998-971b-4b9c91f0a659.png)

vladdeSV:
> Piggy-backing this Issue for still incorrect External Resources

Changed two weeks ago in https://github.com/Zoinkwiz/quest-helper/pull/1071

Mistah210:
Currently shows a link to Night at the Theater for me

vladdeSV:
> Currently shows a link to Night at the Theater for me

From OP's post:

> EDIT: I just determined that the hyperlink being displayed on the Secrets of the North 'tab' will be **whatever quest was viewed/loaded last** in the plug-in.

Issue #1117, opened 2023-04-16
Issue with "External Resources" 


jurrejelle:
When you just open the "Secrets of the North" quest helper, the list of external resources is empty, as expected

If you then open the "Night at the theater" quest helper, it will show the Night at the Theater wiki page as an external resource, as expected

But when you open the "Secrets of the North" helper again, it will still show the Night at the Theater wiki page? [example image showing this](https://cdn.discordapp.com/attachments/855189286575800370/1097069044970049706/image.png)

I assume it has something to do with the external resource list not being cleared if the newest quest  you open doesn't have any external resources, but I haven't looked at the code so not 100% sure.

PR #1120, opened 2023-04-19
Rename Secrets of the North to match external quest resources enum

bryantwong:
Fixes: https://github.com/Zoinkwiz/quest-helper/issues/1080, https://github.com/Zoinkwiz/quest-helper/issues/1117 by matching external quest resources enum: https://github.com/Zoinkwiz/quest-helper/blob/86cd794df5a4d402e8a46aea50c68b545a37808a/src/main/java/com/questhelper/ExternalQuestResources.java#L178

Tested locally:
image


Zoinkwiz (approved):
LGTM, thanks!

PR #1123, opened 2023-04-22
Set The Tourist Trap feather requirement to 50

aivalja:
Strictly only 10 feathers is needed, but the player can lose feathers if creating a dart fails. The wiki recommends 30-50 feathers, so the required amount is set to 50.

### DIFF
--- a/src/main/java/com/questhelper/quests/thetouristtrap/TheTouristTrap.java
+++ b/src/main/java/com/questhelper/quests/thetouristtrap/TheTouristTrap.java
@@ -71,7 +71,7 @@ import net.runelite.api.widgets.WidgetInfo;
 public class TheTouristTrap extends BasicQuestHelper
 {
 	//Items Required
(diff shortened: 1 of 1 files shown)


Felanbird:
Currently your change does nothing, a possible better approach might be to add a tool-tip.

```diff
@@ -233,6 +233,7 @@ public class TheTouristTrap extends BasicQuestHelper
 		prototypeDartTip.setHighlightInInventory(true);
 
 		feather10 = new ItemRequirement("Feather", ItemID.FEATHER, 10);
+		feather10.setTooltip("The wiki recommends 30-50 incase you fail.");
 		bronzeBar = new ItemRequirement("Bronze bar", ItemID.BRONZE_BAR);
 		bronzeBarHighlighted = new ItemRequirement("Bronze bar", ItemID.BRONZE_BAR);
 		bronzeBarHighlighted.setHighlightInInventory(true);
```

Zoinkwiz:
Fixed in 462ee0fb82eee47fd323fb95b969f0d80b585b6d

PR #1129, opened 2023-04-25
Replace imgur links with githubusercontent links

Felanbird:
Next month imgur is [cleaning house](https://help.imgur.com/hc/en-us/articles/14415587638029), which includes deleting an unfortunate amount of anonymously uploaded images.

While I'm unsure if the images used in this repo will be removed, it's very possible, so I went ahead and replaced them now.


Zoinkwiz (approved):
Oh nice, wasn't aware of this! Thanks for the change 😄

Issue #1135, opened 2023-04-30
SOTE: Add stamina potion recommended (at least for library steps)


Paolo-Cosentino:
Just did SOTE over the weekend, noticed the helper doesn't recommend stamina potions. I brought 4 stamina pots with the minimum requirement of 70 agility (it wasn't enough), suggesting at least 6.

PR #1136, opened 2023-04-30
Add stamina potion recommendations to SOTE

Paolo-Cosentino:
Resolves Issue #1135


Obasill:
Why are we using two different variables for stamina potions?

Zoinkwiz (approved):

PR #1142, opened 2023-05-09
Add required items for LunarDiplomacy

tivervac:
There are three points in the quest where you have to fight, but only one of them mentions the combat gear. It's hinted to gather suqah hide, but the step to bring them doesn't mention them as required items.


tivervac:
Still finishing the quest, no need to merge yet

Zoinkwiz (approved):

PR #1155, opened 2023-06-13
Cooks Assistant Rewrite

Obasill:
Was very underwhelmed with the old version especially since this is the first quest that most new players will do. This should 'modernize' the quest and be a lot more approachable for newer players.


Zoinkwiz (approved):
Looks good, thanks for this! As a note I've added a method to make it easier to highlight things such as items in shops interfaces, in this case used to highlight the bucket in the general store.

PR #1160, opened 2023-06-15
Various Fixes

Obasill:
More to come but I wanted to get these in for right now.


src/main/java/com/questhelper/helpers/skills/agility/Agility.java
```diff
@@ -143,43 +139,45 @@ public void setupRequirements()
 	private void setupSteps()
 	{
 		gnomeStronghold = new ObjectStep(this, ObjectID.LOG_BALANCE_23145, new WorldPoint(2474, 3435, 0),
-			"Train agility at the Gnome Stronghold Agility Course.",  Collections.EMPTY_LIST, Collections.singletonList(gracefulOutfit));
```
Zoinkwiz, line 146: The graceful outfit isn't a required item, so is there a reason to change it from recommended to required?

PR #1173, opened 2023-07-04
#1164 death talisman tooltip & alts

### LINKED ISSUE #1164: Catalytic Talisman can be used instead of Death Talisman in ME2 quest
The Catalytic Talisman/Tiara can be used on the death altar in the quest Mournings End. This is very helpful for ironman as its fairly easy to obtain it from the Guardians of the Rift minigame to bypass the 50 items shopping list for those IM who do not have the death talisman.

SpencerSKane:
Added missing alternates to death talisman, cleaned up code a bit with tooltips.
Removed a note that was put as a required item, this note was added to a step description.

I would have liked to only use a single item req, deathTalisman, but deathTalismanHeader remains. It is an important note that should be made aware to the user when checking the initial item req, rather than hidden within a tooltip. I added alts to the header so it wouldn't remain red if you had the alternate items.

closes #1164


Zoinkwiz (approved):
Looks good thanks!

Issue #707, opened 2022-02-06
Murder Mystery Quest OSRS


ijonasb:
The quest is broken on the helper

PR #1175, opened 2023-07-09
2 Murder Mystery Talking Steps: Suspect(typo) & Poison Salesman(bug)

tylerstanczak:
I just did this quest and David was the murderer. However, the NPC Step said "Talk to Anna".

Also, just FYI talking to the poison salesman right before this step does not progress the quest helper log.


tylerstanczak:
fixes #707 hopefully

chestnut1693:
Thank you for resolving this issue. I was experiencing the same problem with the poison salesman today. Looks like this has been an issue for quite some time.

Zoinkwiz (approved):
Thanks for this, hopefully this is the last bug to resolve in the quest!

PR #1233, opened 2023-08-30
Update Underground Pass sidebar item requirements

Fincap:
Bucket and Tinderbox can be obtained from the dwarven camp during the quest when they are needed.

Quest sidebar currently doesn't show this:
image


Felanbird:
The plank can also be obtained during the quest

Zoinkwiz (changes requested):
If we're marking it as obtainable during the quest, we should probably add some checks for 'if you do not have a bucket when you need it, go get the bucket' and such, to avoid people getting lost.

Fincap:
Added in steps to pick up bucket and tinderbox from the dwarven camp.

Zoinkwiz (approved):
Looks good, thanks!

PR #1260, opened 2023-09-11
Removed remaining quantities from item reward names

Qubiz:
As indicated in [this Discord post](https://discord.com/channels/772056816242130964/772056861934354433/1150671194656362537) there were still some occurrences where the quantities in item rewards would be displayed twice. So I didn't catch all the cases in my last PR (#1256).

This PR should fix the remaining occurrences.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ repositories {
 	mavenCentral()
 }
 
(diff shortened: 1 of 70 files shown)


Zoinkwiz (approved):
Thanks :D

PR #1263, opened 2023-09-11
Royal Trouble: Add energy restore item recommendation

pajlada:
Not a very interesting PR - feel free to close if this should be obvious enough for later game quests


Zoinkwiz (approved):
I think having staminas and such always recommended is a good thing these days, if nothing else just as it's nice to have them automatically visible when using the quest bank filter!

PR #1267, opened 2023-09-12
Family Crest: Some polish

pajlada:
- Family Crest: Add teleport hint in the Talk to Caleb step
- Family Crest: Add teleport hint in the Talk to Gem Trader step
- Family Crest: Add teleport hint to the Dwarven Mine
- Family Crest: Add teleport hint to the Witchaven ruin step
- Family Crest: Add teleport hint for smelting the 'perfect' gold bars
- Family Crest: Make use of charged item check for alkharid teleports
- Family Crest: Add "Pull North Lever Again" step to "Avan's piece" category
- Family Crest: Clean up the 'perfect' gold smelting process
- Family Crest: Hint at Varrock tele when heading to Johnathon
- Family Crest: Highlight antipoison in your inventory

Dependent on https://github.com/Zoinkwiz/quest-helper/pull/1268 in its current state, will need a rebase afterwards.
Can also merge that part out if we're not interested in supporting charged items in this way

PR #1268, opened 2023-09-12
feat: ItemRequirement can now check for teleport item charges

pajlada:
This allows for recommended items that calls for 3x Al-Kharid teleports to be fine with letting the user
withdraw 1x Amulet of Glory (3), or 1x Amulet of Glory (6), or 3x Amulet of Glory (1)

This implementation uses data from RuneLite's core ItemCharges plugin,
hence Seth's copyright mention & the link to that file.

https://github.com/Zoinkwiz/quest-helper/pull/1267 shows an example use of this API

<details>
<summary>Video showing the feature off</summary>

[item-charges-test.webm](https://github.com/Zoinkwiz/quest-helper/assets/962989/b999cfbe-f4ff-4273-9463-2abcce8681ed)

</details>


Zoinkwiz (approved):
Thanks this is definitely something that'll help improve things! It'll be good to implement this for infinite teleports (ectophial for example), as well as possibly potions and such in the future, so this'll be a good basis to build that on too!

PR #1271, opened 2023-09-13
Add GitHub Actions CI

pajlada:
This PR adds a simple "gradle build" workflow that will run on commits pushed to the main branch, and to all future PRs

You can see this in action what this looks like on my fork: https://github.com/pajlada/quest-helper/pull/1
the CI job won't automatically run on this PR since the main branch does not have workflow defined, and this PR comes from a fork

If you want to protect your main branch (which would make sure the CI job has to complete before you can merge in PRs), you can:

1. go [here](https://github.com/zoinkwiz/quest-helper/settings/branch_protection_rules/new)
2. set the "Branch name pattern" to `master`
3. enable "Require status checks to pass before merging"
4. type `gradle` in the text box and selecting the job that shows up. if no job has shown up, you might need to have the CI job run in master once first
5. press "Create" at the bottom of the screen

With these settings, you'll always be able to bypass CI if it's ever flaky

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -333,6 +333,7 @@ ij_properties_spaces_around_key_value_delimiter = false
 
 [{*.yaml,*.yml}]
 indent_size = 2
+indent_style = space
 ij_yaml_align_values_properties = do_not_align
 ij_yaml_autoinsert_sequence_marker = true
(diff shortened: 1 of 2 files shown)


Zoinkwiz (approved):
Really nice addition, thanks I'll get this configred!

PR #1286, opened 2023-09-24
Re-working checkRange to make take into account thieving boost

### LINKED ISSUE #1273: Legends' Quest wrong Thieving boost
The "bug" is that it shows as doable at level 47 with "Use Spicy stew for boosts" disabled, because Summer sq'irkjuice gives a 3 boost, but Level 65 Thieving is required to access the summer garden.

The correct boost able level without the Option "Use Spicy stew for boosts" would be at level 48 with the Autumn sq'irkjuice which can be acquired  at Level 45 Thieving.

marmadukeG:
Fixes #1273:

Thieving currently has a static max boost of "3", reworking checkRange to be more dynamic.

### DIFF
--- a/src/main/java/com/questhelper/requirements/player/SkillRequirement.java
+++ b/src/main/java/com/questhelper/requirements/player/SkillRequirement.java
@@ -133,11 +133,15 @@ public class SkillRequirement extends AbstractRequirement
 		{
 			highestBoost = 5;
 		}
-		else if (skill == THIEVING){ //player only has access to Summer sq'irk juice at level 65 thieving which is the default boost value for thieving, currently that's blind to player current skill level
-			if(client.getRealSkillLevel(skill) < 65){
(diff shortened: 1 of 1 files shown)


pajlada (commented):
Small nitpick for the formatting of the solution, otherwise I think this is a good solution for the problem :+1: 

You could format the code slightly different to make it a little bit more clear what it does, like this:
```java
if (...boosts...) {
  ...
}
else if (skill == THIEVING) {
  if (real thieving level < 65) {
    highestBoost = 2;
  } else if (real thieving level < 45) {
    highestBoost = 1;
  }
}
```

Then run IntelliJ's formatter on your new code to align it properly

You can add the text: "Fixes #1273 " in the PR description, that way the issue will automatically close when this PR is merged in!

Zoinkwiz (approved):

PR #1294, opened 2023-09-29
QuestHelperPlugin file refactor

Zoinkwiz:
This PR is intended to help clean up a lot of the bloat inside the QuestHelperPlugin file, delegating instead more to other manager files and reducing replicated code and outdated code (for example for the Cheerer and the menuoptions).

Any feedback on ensuring this is actually implementing decent practices, and what further improvements can be made would be much appreciated!


src/main/java/com/questhelper/managers/QuestBankManager.java
```diff
@@ -0,0 +1,108 @@
...
+	{
+		questBankTab.startUp();
+	}
+
+	public void shutDownQuest()
+	{
+		if (questBankTab != null)
+		{
+			questBankTab.shutDown();
+		}
```
pajlada, line 103: The null-check shouldn't be necessary now since it's a Singleton & injected
```suggestion
		questBankTab.shutDown();
```

src/main/java/com/questhelper/runeliteobjects/extendedruneliteobjects/RuneliteObjectManager.java
```diff
@@ -481,13 +496,9 @@ private void addReplacedNpcOptions(ExtendedRuneliteObject extendedRuneliteObject
 		}
 	}
 
-	public void setActive(ExtendedRuneliteObject extendedRuneliteObject)
+	private void setVisible(ExtendedRuneliteObject extendedRuneliteObject, boolean visible)
```
pajlada, line 499: nit: This could be inlined where it's used, unless there's value in being able to breakpoint here instead of in `ExtendedRuneliteObject`'s `setVisible` function

src/main/java/com/questhelper/runeliteobjects/extendedruneliteobjects/RuneliteObjectManager.java
```diff
@@ -894,7 +956,7 @@ else if (event.getGameState() == GameState.LOADING)
 			runeliteObjectGroups.forEach((groupID, extendedRuneliteObjectGroup) -> {
 				for (ExtendedRuneliteObject extendedRuneliteObject : extendedRuneliteObjectGroup.extendedRuneliteObjects)
 				{
-					disableObject(extendedRuneliteObject);
+					extendedRuneliteObject.render();
```
pajlada, line 959: This feels like a typo to me - previously this would call `disable()` on all objects

pajlada (commented):
Not much feedback on the fake object stuff since I'm not familiar with it, but I found some small nitpicks/questions
and I verified that things build & quest-helper works as expected for cook's assistant

src/main/java/com/questhelper/runeliteobjects/extendedruneliteobjects/RuneliteObjectManager.java
Zoinkwiz: The checks per object have been shifted into the object's person render checks for this, might be worth me changing the naming of render to imply that it checks for hide and show.

PR #1302, opened 2023-10-03
Duke Sucellus -- Fix reversed east & west instructions

C-Swizzle:
image

Notice that the pickaxe is actually on the east, and the pestle and mortar is on the west.
Screenshot has compass with south side up.


Zoinkwiz (approved):
Looks good thanks!

PR #1323, opened 2023-11-04
fix: Make Path of Glouphrie puzzle solver better

### LINKED ISSUE #1317: Path of Glouphrie - Not all combos of values valid for transfer
Currently the helper believes that all combinations of lower-value shapes are valid outputs of the exchanger, so long as they equal the input value. Unfortunately it seems that that's not the case, for example for an input of 10, the only outputs I could produce were:

* Orange square + circle
* Violet circle + red triangle
* Yellow triangle + red circle
* Orange pentagon
* Indigo circle + green circle
* Orange tri + green circle

The helper at the time was trying to get me to get a blue circle/red pentagon, which technically could exist due to 5 + 5, but didn't appear in over 100+ attempts.

At some point we should validate the potential outputs of each value input.

pajlada:
This uses a list of confirmed-by-me exchanges - see code in the linked issue for how this list was achieved.

Fixes #1317


Zoinkwiz (approved):
Looks good from what I can see, I don't have an account I can test it out but I'll trust the results you got, thanks!

PR #1336, opened 2023-11-18
the great brain robbery fix amount of nails needed to repair steps

Elias-Vdw:
make the amount of nails in substep match amount needed


Elias-Vdw:
wait im not actually sure, i think the big requirement might be wrong too. I think this only used 10 nails

Elias-Vdw:
cant really re-do it to make sure...

Elias-Vdw:
the numbers are wrong, but ill abandon this and jsut create an issue i suppose

PR #1357, opened 2023-11-30
Falador Medium Diary: Polish

pajlada:
- The Mogre step now properly highlights the area on the map, and correctly steps through & highlights the Mogre once it has spawned
- Fix the steps not progressing as expected for the Falador wall grapple step
  The coordinates for the wall had changes (or were just wrong from the beginning)
- Change the scarecrow requirement checks to instead ask for the material for a scarecrow, as that's more likely to be what people have.
- Added 47 Farming, 59 Fletching, and 59 Smithing skill requirements for ironmen to grow watermelons & to craft the mithril grapple

I have tested this to the best of my ability on Leagues

### DIFF
--- a/src/main/java/com/questhelper/helpers/achievementdiaries/falador/FaladorMedium.java
+++ b/src/main/java/com/questhelper/helpers/achievementdiaries/falador/FaladorMedium.java
@@ -323,7 +323,7 @@ public class FaladorMedium extends ComplexStateQuestHelper
 		grappleNorthWallStart = new ObjectStep(this, ObjectID.WALL_17050, new WorldPoint(3032, 3389, 0),
 			"Equip your crossbow and grapple then climb the agility shortcut near the Falador Party Room.",
 			anyCrossbow.highlighted(), mithGrapple.highlighted());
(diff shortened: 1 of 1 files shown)


LewisJohnson:
Could you also add bronze spear to "Brain not included" step please?

pajlada:
> Could you also add bronze spear to "Brain not included" step please?

@LewisJohnson Right now the scarecrow items are included as just "Scarecrow", the tooltip includes the actual items for it - including the bronze spear.

Since the bronze spear is the only item we don't have a guide for getting, do you think it should be included as a stand-alone item with the tooltip "only necessary if you don't already have a scarecrow" or something like that?

LewisJohnson:
Scarecrows are non-tradable so really it should have a list of required items for the step in the 

- 1 x Hay sack
- 1 x Bronze spear
- 1 x Watermelon

I don't think many people have a scarecrow chilling in the bank, but that's an opinion not fact.

LewisJohnson:
This PR should still go in, just something for the future

pajlada:
I'm pretty close to it so I think I can just sneak it into this PR before it gets merged :)

pajlada:
This PR is now ready!

PR #1381, opened 2023-12-28
Fix: Flatlaf Runelite compatability

Zoinkwiz:
This makes Quest Helper work with the Runeltie flatlaf changes (https://github.com/runelite/runelite/issues/17284). There's the opportunity still to improve the aesthetic to look better with the new styling (image qualities, some weird outlining, etc), but from what I can tell it's all thankfully fairly functional.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ repositories {
 	mavenCentral()
 }
 
-def runeLiteVersion = 'latest.release'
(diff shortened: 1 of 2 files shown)


pajlada (approved):
I played around with this PR active, started a quest & made sure the panel opened as expected.
As you mentioned things look a bit different, but nothing out of the ordinary or anything that's not illegible

PR #1395, opened 2024-01-16
Add Children of the Sun

Zoinkwiz:


src/main/java/com/questhelper/steps/MultiNpcStep.java
```diff
@@ -0,0 +1,146 @@
...
+import net.runelite.api.NPC;
+import net.runelite.api.coords.WorldPoint;
+import net.runelite.api.events.GameStateChanged;
+import net.runelite.api.events.NpcDespawned;
+import net.runelite.api.events.NpcSpawned;
+import net.runelite.api.events.VarbitChanged;
+import net.runelite.client.eventbus.Subscribe;
+
+
+public class MultiNpcStep extends NpcStep
```
pajlada, line 41: Some basic documentation for the class itself

```suggestion
/**
 * MultiNpcStep should be used over NpcStep when the NPC you're trying to track is a "MultiNPC"
 * TODO: Add some documentation for how you can know an NPC is a MultiNPC, and what resources to use to debug it (e.g. how to find out which varbit controls it, what its base ID is etc)
 */
public class MultiNpcStep extends NpcStep
```

src/main/java/com/questhelper/steps/MultiNpcStep.java
```diff
@@ -0,0 +1,146 @@
...
+{
+	private final List<NPC> baseNPCs = new ArrayList<>();
```
pajlada, line 43:
```suggestion
	/**
	 * List of NPCs matching the base composition ID, but not necessarily the exact NPC ID we're looking for
	 */
	private final List<NPC> baseNPCs = new ArrayList<>();
```

src/main/java/com/questhelper/steps/MultiNpcStep.java
```diff
@@ -0,0 +1,146 @@
...
+
+	private final int multinpcVarbit;
```
pajlada, line 45:
```suggestion
	/**
	 * Varbit that signals a change in the NPC ID, meaning we want to re-checked all suspected NPCs we've found.
	 */
	private final int multinpcVarbit;
```

src/main/java/com/questhelper/steps/MultiNpcStep.java
```diff
@@ -0,0 +1,146 @@
...
+	private final int npcCompositionID;
```
pajlada, line 46:
```suggestion
	/**
	 * The base composition ID of the NPC we're looking for.
	 * @see <a href="https://static.runelite.net/api/runelite-api/net/runelite/api/NPCComposition.html#getId()">NPCComposition#getId()</a>
	 */
	private final int npcCompositionID;
```

pajlada (commented):
The solution looks good to me - I added some basic documentation to the fields & the class itself - not sure if the suggestion will show up as expected

PR #1416, opened 2024-02-21
Add Defender of Varrock

Zoinkwiz:
Fairly bog-standard helper with no new features added.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ repositories {
 	mavenCentral()
 }
 
-def runeLiteVersion = 'latest.release'
+def runeLiteVersion = '1.10.24-SNAPSHOT'
 
 dependencies {
 	compileOnly group: 'net.runelite', name:'client', version: runeLiteVersion
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ repositories {
 	mavenCentral()
 }
 
-def runeLiteVersion = '1.10.24-SNAPSHOT'
+def runeLiteVersion = 'latest.release'
 
 dependencies {
 	compileOnly group: 'net.runelite', name:'client', version: runeLiteVersion
--- a/src/main/java/com/questhelper/helpers/miniquests/familypest/FamilyPest.java
+++ b/src/main/java/com/questhelper/helpers/miniquests/familypest/FamilyPest.java
@@ -110,7 +110,7 @@ public class FamilyPest extends BasicQuestHelper
 
 	public void setupSteps()
 	{
-		talkToDimintheis = new NpcStep(this, NpcID.DIMINTHEIS, new WorldPoint(3280, 3404, 0), "Talk to Dimintheis (Southeast of Varrock.");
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -10,7 +10,7 @@ repositories {
 	mavenCentral()
 }
 
-def runeLiteVersion = 'latest.release'
+def runeLiteVersion = '1.10.24-SNAPSHOT'
```
pajlada, line 13: This should be reverted to `latest.release` as the snapshot is no longer necessary

pajlada (commented):
## General feedback

- The quest is missing the list of enemies to defeat
From the wiki:
At least six [armoured zombies](https://oldschool.runescape.wiki/w/Armoured_zombie) (level 85)
Multiple [Chaos Golems](https://oldschool.runescape.wiki/w/Chaos_Golem) (level 70)
- The teleports used should be part of the "Recommended items" list

## Step feedback

If cancelling out of the cutscene with Arrav, we should be able to tell the user to click the gate since they're past the gate. So something along the lines of:
`goSeeArrav.addStep(inSecondSectionOfTheDungeon, clickTheDoorAgainAndWatchTheCutscene);`

The "Kill zombies to fill 3 bottles with mist again and continue deeper into the dungeon" should probably highlight bottles if you don't have 3 bottles, and it would be nice if it also highlighted the red mist that you must collect

For the "Talk to Captain Rovin upstairs in the north west of Varrock Castle" step
1) Maybe suggest a Varrock Teleport
2) Highlight the "Climb up" on the middle stairs

For the "Talk to Reldo in the Varrock Castle library" step
1) Highlight stairs to go down to the bottom floor

The "Talk to Curator Haig in the Varrock Museum" step was skipped for me

pajlada:
The quest guide is missing quest & skill requirements
image

pajlada (commented):
If you cancel out of the dialog during the "Talk to Elias in the dungeon." step at the beginning of the cave, you won't be able to look over the balcony because Elias isn't following you.
elias-left-behind
You can check that Elias is following you with [Varp 447](https://chisel.weirdgloop.org/varbs/display?varplayer=447).
-1 denotes no follower, while the value is set to `835887551` when Elias is following me, which gives me the npc id of 12754 (although his NPC ID is 12610).
We can either be safe and just ensure you have a follower at all (check if it's something other than -1), or check that the highest 16 bits are 12754 or 12610
https://oldschool.runescape.wiki/w/RuneScape:Varplayer/447 has more info about the varp

---

If someone else mined the same rock, no other rocks are highlighted
no-highlight-cuz-someone-else-mining

---

Missing varrock teleport recommendation when returning to the Varrock Castle
no-tp-recommendation

---

No dialog highlight when attempting to climb up the stairs
no-climb-up-recommendation

---

When prompted to go to Reldo in the library, the stairs aren't highlighted for you to go down
no-stair-highlight

---

When finally talking to Dimintheis, depending on your overall quest state, you might need to select "Other" first. "Other" should be added as a dialog highlight for this step
should-highlight-other-with-diminethis

PR #1434, opened 2024-02-24
feature: Added pole obtaining step to MD

### LINKED ISSUE #611: Mountain Daughter broken pole ItemRequirement
[The [pole] requirement ](https://github.com/Zoinkwiz/quest-helper/blob/86fc0a905f2086678652595da3dbb14c9415be19/src/main/java/com/questhelper/quests/mountaindaughter/MountainDaughter.java#L178) is not working which results in reaching `poleVaultRocks` without any means of completing it. 

image

Additionally, it might be helpful to add a `step` rather a note to pick up the pole on the ground nearby as it seems extremely unlikely that anyone would be carrying around a [staff](https://oldschool.runescape.wiki/w/Staff) (and other, commonly used, staves -I tested a [Smoke Staff](https://oldschool.runescape.wiki/w/Mystic_smoke_staff)- don't seem to work)

Zoinkwiz:
closes #611


pajlada (approved):
Not tested in-game but seems reasonable

PR #1444, opened 2024-02-26
Add puzzle solver toggle

Zoinkwiz:
This PR intends to add a togglable option to not display puzzle solutions whilst using Quest Helper. The line between puzzle and not puzzle can be a bit blurry, so any feedback on the usage of it and lack of usage in places is appreciated.


pajlada (approved):
Tested this on Ernest the Chicken and it works as expected :+1: 

I tried toggling it off & on a few times during the puzzle itself & it always kept up-to-date

PR #1447, opened 2024-02-27
fix:Allow for decrementing state of quests

### LINKED ISSUE #1221: Demon Slayer stuck.
Demon slayer is stuck thinking I have completed Rovins and Prysins keys when I havent even started the quest yet. This means it also doesn't remember the correct incantation as it thinks it already has seen it.

Zoinkwiz:
Usually a quest doesn't have a decrementing value for its state, however for Witch's House and on Quest Speedrunning worlds it can.

This allows for detection of changing state even on the value going down.

fixes #1221


pajlada:
Tested this on a speedrunning world & when resetting a quest it worked as expected.
The "lockable" categories stayed locked, but that didn't change the behaviour of the quest steps

image

PR #1452, opened 2024-02-27
feat:Added opacity options on colour highlights

Zoinkwiz:


pajlada (commented):
The opacity value doesn't have any limits, seems like I can set it to any integer above 0. It should probably be limited to 0-255 (or 1-255) if those are the expected values

It only works with "Square" item highlights & "Convex Hull" NPC highlights, is this intended?

Zoinkwiz:
Good point on the range, added now! On the highlight choice, you're right in that I think there are some for widgets and such I should check out as well, I'll look into that now.

Zoinkwiz:
I realised the existing color config could be configured to also accept opacity as an option, so I've adjusted the PR to try and make use of that, as well as making some useful re-usable functions for the common usage of various opacities.

PR #1473, opened 2024-03-04
fix:Handle incorrect state assumption in AKD

### LINKED ISSUE #697: A kingdom divide
After speaking with all the leaders at the xeric tower, helping the arceeus with the shield potion, the "speak with all the leaders again on each floor of xeric's lookout" and the user interface is pointing Arceeus leader, nothing happens with any of  them, I have to carry out the deeds for all the other house leaders

Zoinkwiz:
I haven't been able to test it, but this should handle a scenario where the player can talk to one of the leaders of Kourend before finishing all of their tasks, which'd progress the state. This wasn't then handled correctly.

fixes #697


pajlada (approved):
The only thing I could think of to make this more safe without testing is to replace the fallback `new DetailedQuestStep(this, "You should not see this!")` of `megaStep` with `talkToAllLeadersLookoutFinish`, but might not be worth the restructuring - leaving that decision up to you!

PR #1506, opened 2024-03-21
feat:Add Twilight's Promise

Zoinkwiz:
This adds Twilight's Promise, as well as the base template for the remaining 3 new quests.


pajlada (commented):
First step phrasing, I'd prefer "Talk to Regulus Cento outside Varrock's East Gate to travel to Varlamore" or something along those lines
After returning to Ennius Tullus, I was prompted to talk to the knight in the Bazaar first, but that step is the 4th of the knight steps. If this is not a per-person thing, we should move the Bazaar step to the top. My order was: Bazaar -> Cothon -> Pub -> Colosseum

Maybe highlight the entrance to the Colosseum?
image

When prompted to return to Ennius Tullus, maybe highlight stairs up from colosseum if the user is inside?
image

Stairs down were not highlighted
image
image
This might just be because it's not possible for these stairs, I know some things just can't be highlighted like this

Maybe rename the "Cult" category to "The Twilight Emissaries" (from wiki)

PR #1508, opened 2024-03-22
Added Ribbiting Tale

Zoinkwiz:


pajlada (approved):
Great quest, great helper. Loved the category names. 10/10

PR #1512, opened 2024-03-23
Children of the sun very small edit

Haavardaw:
Just removed a part where it said When varlamore releases.

### DIFF
--- a/src/main/java/com/questhelper/helpers/quests/childrenofthesun/ChildrenOfTheSun.java
+++ b/src/main/java/com/questhelper/helpers/quests/childrenofthesun/ChildrenOfTheSun.java
@@ -265,7 +265,7 @@ public class ChildrenOfTheSun extends BasicQuestHelper
 	public List<UnlockReward> getUnlockRewards()
 	{
 		return List.of(
-			new UnlockReward("Access to Varlamore (when it's released)")
+			new UnlockReward("Access to Varlamore to the east of Varrock with a bird.")
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/questhelper/helpers/quests/childrenofthesun/ChildrenOfTheSun.java
```diff
@@ -265,7 +265,7 @@ public QuestPointReward getQuestPointReward()
 	public List<UnlockReward> getUnlockRewards()
 	{
 		return List.of(
-			new UnlockReward("Access to Varlamore (when it's released)")
+			new UnlockReward("Access to Varlamore to the east of Varrock with a Quetzal.")
```
pajlada, line 268: nit: To keep consistent with other area unlocks, & the quest completion picture, I'd leave this as just "Access to Varlamore"

Haavardaw: That is fair, I figured giving location as the main reason most people do this quest is only to get access to Varlamore and I saw so many questions about how to get there on release. Maybe we add something for when you start each quest if you are not already in Varlamore when starting the helper or anything along those lines, not sure what you think personally. Or somewhere else?

Haavardaw: Yep, could possibly add just a check for if you're within Varlamore or not when opening guide. But for now I'll just send a new PR when I do changes to other stuff with this one.

Kerpackie: > Yep, could possibly add just a check for if you're within Varlamore or not when opening guide. But for now I'll just send a new PR when I do changes to other stuff with this one.

Best approach, do a zone check wrapping all of Varlamore in it, and having it as the first step. Then if they aren't there, they are presented with a step for "Go to Varlamore. -> "Talk to Regulus Cento outside Varrock's East Gate to travel to Varlamore." like Paj recommended.

We integrate a similar step then to all quests that start on Varlamore, could do similar for Zeah/Fossil Island which generally have their primary mode of transport to the area being an interaction. (And add teleport options as applicable using the current recommended teleport system).

PR #1514, opened 2024-03-28
feat: Added Xeric's Talisman tele suggestions to A Kingdom Divided

HairyDan:
All steps are completable with Kharedst's Memoirs, but often Xeric's Talisman is a little faster - suggesting both here.

One thing I would have liked to add is a highlight on which option to take on the Xeric's Talisman - but afaict that's not possible (at least it's not a dialogue step)

image


Zoinkwiz (approved):
Looks good, thanks for this!

PR #1524, opened 2024-04-06
Add eagles peak as requirement to at first light

AdressaAgne:
The quest At first light does require the Eagles Peak quest


pajlada:
Duplicate of #1520

PR #1541, opened 2024-04-27
Herb run varlamore

### LINKED ISSUE #1532: Add new Varlamore patch for the Herb Run guide
With the new Varlamore update, there's a new farming patch with an herb, flower, and two allotment patches. The herb guide should have this added and the respective teleports, number of seeds needed, and possible teleport options would be greatly helpful.

Thank you!

SpencerSKane:
Added the Varlamore herb patch to the generic helper

Added each variant of the whistle as suggested teleport item
closes #1532


Zoinkwiz (approved):
Looks good, thanks for this!

PR #1542, opened 2024-04-28
feat: Warn on failed varbit reads

pajlada:
Previously, if a varbit failed to read (e.g. if the varbit had been deleted), we didn't handle the exception causing the helper & client to misbehave.
Now, we catch the exception & post a message in chat instructing the user to notify us via the Quest Helper discord.
To be able to achieve this, I have changed the signature of `check` to contain the `ChatMessageManager`

To test this, open the Kourend medium diary helper & skip the steps until you get to the bluegill step, it should show up like the image below

image

Questions:
 - Is the color for the chat message ok?


pajlada:
Closing in favour of #1543

PR #1543, opened 2024-04-28
feat: Warn on failed varbit reads (without big refactor)

pajlada:
Previously, if a varbit failed to read (e.g. if the varbit had been deleted), we didn't handle the exception causing the helper & client to misbehave.
Now, we catch the exception & post a message in chat instructing the user to notify us via the Quest Helper discord.
This uses `client.addChatMessage` which can only be called from the client thread. The varbit check function is only ever called from there, so for this use this is ok.

image

To test this, open the Kourend medium diary helper & skip the steps until you get to the bluegill step, it should show up like the image below

Questions:
 - Is the color for the chat message ok?


Zoinkwiz (approved):
Looks good and nice and clear, thanks!

Issue #1550, opened 2024-05-03
Minor bug - Draynor Village


ScullyB96:
Not really a bug, but having a charged glory does not trigger the green text as it relates to a Draynor teleport.

Zoinkwiz:
Hi @ScullyB96 , for which quest did you have this issue?

SpencerSKane:
#1551 

I searched the quests and was able to find two that could use updates in regards to the Draynor tele. This potentially fixes the issue but can't be certain as we don't know the quest this was encountered on.

PR #1551, opened 2024-05-03
cleared up some draynor tele options

SpencerSKane:
Added draynor tab and glory alts in quests where there were issues


SpencerSKane:
fixes #1550

Zoinkwiz (approved):

PR #1556, opened 2024-05-11
fix: retain opacity on bank tab reopen

### LINKED ISSUE #1547: [Bug] Missing items opacity in bank filter doesn't apply on bank reopen
When reopening the bank with the bank filter on, the missing items are no longer greyed out. You can see in the screenshots with the eternal glory and the air staff. Toggling the filter off then on again resets the opacity back to what it should be.
image
image

adamk33n3r:
Fixes #1547


pajlada (approved):
Confirming this fixes the bug for me :+1:

PR #1558, opened 2024-05-14
feat: Made the Memory Puzzle in Lunar Diplomacy dynamic

Zoinkwiz:
This adjusts the path generation from being hardcoded options to dynamic based on the actual path. This should ensure it's more reliable and less prone to error.

### DIFF
--- a/src/test/java/com/questhelper/helpers/quests/lunardiplomacy/MemoryChallengeTest.java
+++ b/src/test/java/com/questhelper/helpers/quests/lunardiplomacy/MemoryChallengeTest.java
@@ -92,7 +92,7 @@ public class MemoryChallengeTest
 		return MemoryChallenge.findPath(grid);
 	}
 
-	private int[][] createGrid(int west, int middleWest, int middleEast, int east)
+	int[][] createGrid(int west, int middleWest, int middleEast, int east)
 	{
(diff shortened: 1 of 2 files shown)


src/test/java/com/questhelper/helpers/quests/lunardiplomacy/MemoryChallengeTest.java
```diff
@@ -0,0 +1,109 @@
...
+		assertEquals(10, path.size(), "Path should consist of 10 points including start and end ledges.");
+	}
+
+	private List<WorldPoint> runPathTest(int west, int middleWest, int middleEast, int east)
+	{
+		int[][] grid = createGrid(west, middleWest, middleEast, east);
+		return MemoryChallenge.findPath(grid);
+	}
+
+	private int[][] createGrid(int west, int middleWest, int middleEast, int east)
```
pajlada, line 95: You could make MemoryChallenge's package-private instead (just remove `private`), that way the test can use MemoryChallenge's `createGrid` since they're in the same package.
There's some VisibleForTesting stuff but I'm not sure how that works with functions

pajlada (approved):
Code looks sane to me, haven't tested.
One optional nit for the unit test

PR #1559, opened 2024-05-14
Remove Gnome Stronghold from Agility guide

Zoinkwiz:
This is due to Draynor course becoming a level 1 course.


pajlada (approved):
LGTM :+1: 
Only thing I notice is that the category step thing doesn't get highlighted/expanded by itself, but I suspect that's not due to any changes you've done

image

PR #1569, opened 2024-05-19
feat: list numbers of active skill filters next to the dropdown

pajlada:
I've noticed some support tickets coming in regarding quests not being visible, and often the cause is that the user has accidentally, or just forgotten, that they've enabled some skill filter.

This PR aims to make it easier both for users & for support members to identify this by showing how many (if any) skill filters are enabled next to the "Skill filtering" setting

### No skill filter active, menu collapsed

image

### No skill filter active, menu expanded

image

### 3 skill filters active, menu collapsed

image

### 3 skill filters active, menu expanded

image


Zoinkwiz (approved):
This is a really clean way of making this clearer, thanks for this!

PR #1574, opened 2024-05-24
Remove event name changing in the QuestBankTab

Zoinkwiz:
Tried wracking my brain for why this was important when I added it, but even the comment doesn't help me lol. Everything 'seems' to work properly without it, and it avoids any potential future issues. I have a hunch it might've been something to do with the quest filtering functionality sometimes getting overrided by a normal bank tab, but I've not seen that happen without it 🤷 .

It'd be great if I could get a second pair of eyes on it just to make sure nothing is out of place.


pajlada (approved):
Everything seems to be working as expected for me
The selected quest helper tag is still selected after selecting it, closing the bank, and opening the bank up again
The selected quest helper tag is deselected when closing down a quest helper
The quest helper tag shows up in the interface again when starting a quest helper
I can switching between bank tags, bank tabs, and the quest helper tag, as expected

PR #1575, opened 2024-05-24
Remove duplicate subscribes on overridden functionality

Zoinkwiz:
A function with Subscribe which overrides another function with Subscribe will mean there are two invocations of the function. This removes occurrences of this.


pajlada (approved):
Confirmed this works as expected (only fires once now instead of twice). Would be nice if there was a lint to stop this from happening :shipit:

PR #1578, opened 2024-06-02
feat: highlight spell widget based on spell name

pajlada:
I went with the OOP approach of making the existing `WidgetHighlights` into an abstract class (now `AbstractWidgetHighlight`), moving the old implementation to `WidgetHighlight`.

`SpellWidgetHighlight` extends `AbstractWidgetHighlight` and adds the `addSpellHighlight` function to quest steps.

Example usage:
```java
castNormalSpellbookSpell = new ObjectStep(this, ObjectID.ANCIENT_GATE_2930, new WorldPoint(2763, 9314, 0), "Cast a charge orb spell on the ancient gate.", normalSpellbook);
castNormalSpellbookSpell.addSpellHighlight("Camelot Teleport");
```

The implementation of `SpellWidgetHighlight` is stateful in its approach to ensure we don't look for the spell more than necessary.
Right now, a spell search only takes place once every time the spellbook is opened. Theoretically, this might not even be necessary, as we might just need to do it once in total.
https://github.com/Zoinkwiz/quest-helper/pull/1578/files#diff-1e58a657a71bcd5ce13d4f0cd7a6657f49e5acc831ce50d45c2a8197876a9f6eR72 could be removed to only search once, but I have not tried that out

Every time the widget is highlighted, I confirm the spell name is actually what we're looking for https://github.com/Zoinkwiz/quest-helper/pull/1578/files#diff-1e58a657a71bcd5ce13d4f0cd7a6657f49e5acc831ce50d45c2a8197876a9f6eR104-R111 in case the component id is reused for something else. I never noticed this failing, but I thought I'd leave it here to rather be safe than sorry.

There are one potentially sticky portions: We hardcode the child ID of the spell list, since it's not something RuneLite themselves use we won't get this auto updated.

### DIFF
--- a/src/main/java/com/questhelper/steps/AbstractWidgetHighlight.java
+++ b/src/main/java/com/questhelper/steps/AbstractWidgetHighlight.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2020, Zoinkwiz
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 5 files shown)


Zoinkwiz (approved):
Looks good, all seems to work as expected!

PR #1579, opened 2024-06-04
feat: Added Spell enum for spells

Zoinkwiz:
This is intended for usage with the SpellWidgetHighlight, as an easy way to specifiy the spell you want to use.


pajlada (approved):
Only thing I'd consider is removing the raw String constructor for `SpellWidgetHighlight`, otherwise seems good

PR #1580, opened 2024-06-04
Add Bike Shedder development quest

pajlada:
this quest is intended to be a place where example usages of steps/requirements/conditions can be added and tests in quick succession
current steps are just there as an example, and take place in lumbridge


Zoinkwiz (approved):
Looks good and seems to work as expected for dev mode. Shame I can't test the lightbearer step 😂

PR #1586, opened 2024-06-06
Add Tree Run Helper

Zoinkwiz:
This is a helper to hopefully provide some nice utility such as the auto-bank for tree runs, as well as help keep make sure you don't accidentally miss out on paying for protection. There's various toggles for preference. 

I think it could use further guidance, and really could benefit from some way to move about the order depending on preference. I think that needs to be a project in its own right though.

### DIFF
--- a/src/main/java/com/questhelper/bank/banktab/QuestHelperBankTagService.java
+++ b/src/main/java/com/questhelper/bank/banktab/QuestHelperBankTagService.java
@@ -150,9 +150,18 @@ public class QuestHelperBankTagService
 		{
 			BankTabItems allRequiredItems = new BankTabItems("Required items");
 			List<ItemRequirement> allRequired = plugin.getSelectedQuest().getItemRequirements();
+			List<ItemRequirement> items;
 			if (allRequired != null && allRequired.size() > 0)
 			{
-				allRequired.forEach(item -> getItemsFromRequirement(allRequiredItems.getItems(), item, item));
+				items = allRequired.stream()
+					.filter(Objects::nonNull)
+					.map(ItemRequirement.class::cast)
+					.filter(i -> (!onlyGetMissingItems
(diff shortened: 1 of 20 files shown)


Kerpackie:
Thanks for taking the time out to finish this for me, great to see it finally completed!

pajlada:
Does not recommend a rake, although maybe that's correct since users should know about raking patches by this point and would just clutter the guide up

The Gnome Stronghold step skipped the normal tree, and only did the fruit tree for me

The dialog steps for chopping a tree down is missing the slow dialog steps, they should be
`step.addDialogSteps("Would you chop my tree down for me?","I can't be bothered - I'd rather pay you to do it.", "Here's 200 Coins - chop my tree down please.", "Yes.");`

Would be nice if Elkoy was highlighted to guide you out (i know the guide tells you)

When I did the fossil island hard trees, it told me I was done after just planting one of the 3 trees

Kerpackie:
> When I did the fossil island hard trees, it told me I was done after just planting one of the 3 trees

This could be an issue with how the varb for the trees updates. There is a shared varb for a bunch of patches, and it gets updated based on how close you are to a specific patch. Considering how close the fossil island plots are, if they are all sharing the same varb then the zoning will need to be really specific for it or it would skip the step.

Brandon1212b:
From this thread it sounds like there are tree runs in quest helper but I am not seeing it currently. Is there any way I can add this?Im not too familiar with Github, I am sorry if this is a noob question.

kcjg94:
> From this thread it sounds like there are tree runs in quest helper but I am not seeing it currently. Is there any way I can add this?Im not too familiar with Github, I am sorry if this is a noob question.

It is currently in development. The tree run helper is not on the version of Quest Helper available in the plugin hub through Runelite.

PR #1590, opened 2024-06-09
Assert in requirements/conditions, ensuring values are not set to null

pajlada:


pajlada:
I applied your fixes from https://github.com/Zoinkwiz/quest-helper/pull/1591

Zoinkwiz (approved):
This is great, thanks!

PR #1594, opened 2024-06-11
Fix widget ids for garden of tranquility diplomat quiz

nea89o:
Adjusts the widget ids for the diplomat quiz during garden of tranquility

image


Zoinkwiz (approved):
Thanks for this, looks good!

PR #1598, opened 2024-06-15
fix: respect "Display icons on NPC and objects" settings when prompted to use item on object

### LINKED ISSUE #1502: Add the ability to hide the "use item" highlight on top of objects
From Discord https://discord.com/channels/772056816242130964/776427467560321034/1219394126605320242

The feedback is that this icon 
image
can be distracting when the object you need to click is small

pajlada:
This changes the behaviour of the setting **Display icons on NPC and objects** when prompted to use an item on an object

Setting **on**, both prior & with this PR, when prompted to use something on an NPC
a2

Setting **off**, both prior & with this PR, when prompted to use something on an NPC
a1

Setting **on**, both prior & with this PR when prompted to use an item on an object
b1

Setting **off**, with the changes of this PR when prompted to use an item on an object (**This is the new behaviour**)
b2

Fixes #1502

### DIFF
--- a/src/main/java/com/questhelper/playerquests/bikeshedder/BikeShedder.java
+++ b/src/main/java/com/questhelper/playerquests/bikeshedder/BikeShedder.java
@@ -25,6 +25,7 @@
 package com.questhelper.playerquests.bikeshedder;
 
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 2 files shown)


Zoinkwiz (approved):
Looks good, thanks!

PR #1634, opened 2024-07-12
Use any crossbow for ThePathOfGlouphrie.

mrhappyasthma:
The list was manually adding alternatives and missed the dragon hunter crossbow. Rather than appending it manually, I opted to reuse the `ItemCollections.CROSSBOWS`.


pajlada (approved):
Seems good - the only crossbow not able to be used for a grapple is the Love Crossbow, which isn't in the `ItemCollections.CROSSBOWS`

Zoinkwiz (approved):

PR #1652, opened 2024-07-24
While guthix sleeps

Zoinkwiz:


pajlada (changes requested):
I've made a PR changing a few tiny things that I couldn't really explain in text, or I just happened to make while running it: https://github.com/Zoinkwiz/quest-helper/pull/1670

The version change to `build.gradle` should be undone

## Starting off

For required items, say "Combat gear" instead of just "Melee weapon". Graceful + whip did not work super well, especially when being bad at prayer flicking.

For recommended items, recommend more than just 1x food.

For the "Kill the assassins" step, you should camp pray range, and switch to pray melee right after the arrow hits you, then switch back to pray ranged after the DDS spec (not dragon claw spec) on you.

## Investigating

For the "Talk to the Hunting Expert in his hut in the middle of the Feldip Hills hunting area" step and "Bring the unconcscious broav...", the dialogue option for me was "A creature to help track down a person."

img

## Movario's Base

For required items, the runes should be marked as "Can be obtained during the quest"

The "Search the open chest in the far north of the area for some runes" is not part of the Movario's Base category, so your progress gets lost in the side panel. It should be added under the "Enter movario's base", and above "Climb down the stairs to the north-west".

nit: Remove `that've` word, so maybe "Climb up the stairs that appeared" instead of "Climb up the stairs that've appeared"

When at the weight puzzle, if I go upstairs, it highlights the "Search the desk north-west of the staircase" step in the side panel

## United Front

The "Return to Thaerisk" step should recommend you use a taverley teleport. Earlier steps should recommend bringing a taverley teleport too

After the "Return to Thaerisk step", so the "Kill the mercenaries step", we can tell the user that they can destroy the movario's notes from their inventory.

It tells me I have to be on Lunar spellbook, then asks me to use a seer's village teleport. Does it ask for the lunar spellbook a bit early? maybe worth to split up this category, or prompt the user to bring teleport tablets

The steps have teleport recommendations that aren't part of the categories recommended items

Missing 20 coins requirement for the betty step

nit :The "Plant the enriched snapdragon seed" should have "Yes." as a dialog step

"Talk to Ghommal outside of the entrance to the warrio's guild" can have a teleport requirement

## Infiltration

As a requirement, you should need to be on normal spellbook

You need ~9 free inventory slots to pick up all the gear, this should be recommended before the users heads off to the black knight castle

There was a decent amount of running, so unless the user is bringing graceful, a stamina would be nice for the running part. Maybe recommend using protection prayers too, they slapped me pretty hard

When the user needs the death rune/law rune, ensure the user actually brings them by adding them as step item requirements (unless the user is able to pick up an infinite amount of death/law runes in there)

## ???

Maybe tell user they can drop spade/hammer/chisel once they are done with them?

Maybe rename this category?

#### Herb puzzle

nit: Would be nice to highlight the druid pouch & the druids for when you need to get the ingredients from them. But for it to be really nice, we'd need to check if the ingredients are already on the floor. Maybe overthinking

Would be nice to use the "line points" for the pathing to the various herb altars, since the explanations aren't super good. I don't know how to make the explanations better either

The "Cast bloom" step is not part of the category, so category focus is lost when you reach that step

I would split up the "???" category and "clicking the orb of jas" categories personally, so the player doesn't gear up in full torva for the running/herb puzzle part

"Talk to Idria to teleport back to Falador Castle" can have "Yes." added as a dialog step

pajlada:
Random things to check as a last-second check:

1. Is the difficulty correct?
2. Are the skill requirements correct?
3. Is the position in the optimal quest guide & optimal ironman quest guide correct?
4. Are the monsters you have to fight correctly listed?
5. Are the rewards correct?
6. Does the "RuneScape Wiki" link work?

pajlada (approved):

PR #1655, opened 2024-07-24
fix: don't show empty list of require items in side panel

pajlada:
This PR refactors a good chunk of the `QuestOverviewPanel`

Requirements in steps or general requirements are now handled by the `QuestRequirementsPanel` which stores the list of requirements and can know whether or not to hide itself if no requirement is visible

In most places where we've previously used a `JLabel` + html to do formatting/text wrapping/styling, it's now a `JTextArea` with input disabled

I've tried to make the "Open RuneScape Wiki" text look a bit more readable when you hover over it

| before | after |
| ---- | ---- |
| before | after |

it does a few more things than I'd like, if you prefer I can split it up into a few smaller PRs instead


Zoinkwiz (approved):
Seems good and seems to work from what I've tested, thanks as always!

PR #1671, opened 2024-08-04
AllNeededItems: specify in tooltip that "Highlight missing items" behavior is a Quest Helper config setting

cdfisher:
This tooltip has definitely cut back of the number of messages in RuneLite support related to "Highlight missing items" being enabled, but there is still occasional traffic about it.

This PR changes the message to say that it's specifically a Quest Helper setting, rather than just saying it's a config setting.


Zoinkwiz:
Already implemented with https://github.com/Zoinkwiz/quest-helper/commit/cebd56c41852a420671422f0a44128c87ffb99fa

PR #1675, opened 2024-08-05
WGS Feedback final round

Haavardaw:
I already mentioned most the stuff I changed to Zoinkwiz but have a look around it. Stuff like adding General requirements

### DIFF
--- a/.run/Application.run.xml
+++ b/.run/Application.run.xml
@@ -1,6 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Application" type="Application" factoryName="Application" singleton="false">
-    <option name="ALTERNATIVE_JRE_PATH" value="BUNDLED" />
+    <option name="ALTERNATIVE_JRE_PATH" value="15" />
     <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="true" />
     <option name="MAIN_CLASS_NAME" value="com.questhelper.QuestHelperPluginTest" />
     <module name="quest-helper.test" />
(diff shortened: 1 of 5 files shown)


.run/Application.run.xml
```diff
@@ -1,6 +1,6 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Application" type="Application" factoryName="Application" singleton="false">
-    <option name="ALTERNATIVE_JRE_PATH" value="15" />
+    <option name="ALTERNATIVE_JRE_PATH" value="BUNDLED" />
```
pajlada, line 3: should probably undo changes to this file, at least as part of this PR

src/main/java/com/questhelper/panel/questorders/QuestOrders.java
```diff
@@ -48,9 +48,9 @@ public class QuestOrders
 		QuestHelperQuest.IMP_CATCHER,
 		QuestHelperQuest.WITCHS_POTION,
 		QuestHelperQuest.GERTRUDES_CAT,
+		QuestHelperQuest.CHILDREN_OF_THE_SUN,
```
pajlada, line 51: would be nice to undo all quest order changes for this pr, and have that as a separate PR instead, just for commit sanity once it lands in the main branch

pajlada (changes requested):
Some nits about unrelated changes, otherwise seems good :+1: 
Once this PR lands into the `while-guthix-sleeps` branch, Zoinkwiz should run the formatter on it too since I see some formatting errors in your changes

PR #1680, opened 2024-08-07
fix: Corrected west->east order in Lunar Diplomacy Memory Puzzle

Zoinkwiz:
I think this is right, looking at what the values stored were for the varbits used for the columns of the memory puzzle. However, ideally we do test this first.


umer-rs:
Looks good now in game.

PR #1713, opened 2024-08-22
persist the selected quest upon relogging

umer-rs:
reopens the previously active quest upon logging out. 

added a new boolean flag because we need to differentiate between the runescape profile changing and being logged out. we can be logged out but be logging in to the same runescape profile.

case 1: log out and log into same runescape profile -> reopen previous quest

case 2: log out and log into different runescape profile -> dont open previous quest

case 3: don't log out but hop to a different runescape profile -> dont open previous quest

could also potentially add a config option for this.


kcjg94:
I'm not sure if this should be the default (i.e. it should require a config option). If a user is doing a quest, and then moves to a different activity, on their next login it shouldn't be opening the quest helper plugin. It seems a bit invasive of the quest helper plugin to move from one plugin to itself.

PR #1722, opened 2024-09-01
fix: Fixed Guthix rest checks for WGS

Zoinkwiz:
The overarching condition already has NOR in it, so by noring inside a NOR'd Condition it was presumably negating itself. I hope this fixes it.

Hopefully we can test this though before merging though.


Zoinkwiz:
Tested and seem to work.

PR #1726, opened 2024-09-03
Feat add assistance level groupings

Zoinkwiz:
This PR is intended to make it easier for players to get started with Quest Helper in a customised to desired use way. It adds three tiers of assistance, minimum, medium, and full assistance.

When you first click on a quest after this update is live in the sidebar, you will be prompted to select an assistance level.

Screenshot 2024-09-03 141422

When you select one, you will be taken to the normal quest page.

If you hit the cog at the top of the sidebar, it will let you access this page again to adjust the level of assistance you're getting.

Currently this prompt won't occur if you activate the helper via the in-game quest list, or if the quest helper is activated automatically. I think we may want these both captured, but I want to make sure we don't result in more confusion from people say not knowing about the sidebar, and then not being sure why their helper hasn't started up.

### DIFF
--- a/src/main/java/com/questhelper/panel/AssistLevelPanel.java
+++ b/src/main/java/com/questhelper/panel/AssistLevelPanel.java
@@ -1,0 +1,196 @@
+/*
+ * Copyright (c) 2024, Zoinkwiz <https://github.com/Zoinkwiz>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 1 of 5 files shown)


src/main/java/com/questhelper/panel/AssistLevelPanel.java
```diff
@@ -0,0 +1,196 @@
...
+		minAssistPanel.setLayout(new BorderLayout());
+
+		JLabel title = new JLabel();
+		title.setText("<html>Minimum Assistance</html>");
+		title.setForeground(Color.WHITE);
+		title.setBorder(BorderFactory.createEmptyBorder(10, 10, 0, 0));
+
+		JLabel description = new JLabel();
+		description.setText("<html>This disables most overlay-based help, and hides puzzle solutions. This is a good option if you just want the sidebar with helpful information, as it will still track where you are in the quest, and show items needed.</html>");
+		description.setBorder(BorderFactory.createEmptyBorder(10, 10, 10, 10));
```
pajlada, line 105: Not a requirement by any means, but with this function added in Utils.java:
```java

	public JTextArea createStaticTextArea(String text)
	{
		var textArea = new JTextArea();
		textArea.setLineWrap(true);
		textArea.setWrapStyleWord(true);
		textArea.setOpaque(false);
		textArea.setEditable(false);
		textArea.setFocusable(false);
		textArea.setText(text);
		return textArea;
	}
```

You could replace this with
```suggestion
		var description = Utils.createStaticTextArea("Gives you the full helper experience, highlighting objects & dialogue options, puzzler solvers, and with arrows so you know where to go.");
		description.setBorder(BorderFactory.createEmptyBorder(10, 10, 10, 10));
```

removing the need to use HTML to get text wrapping

pajlada (commented):
With some mini testing, this seems to work as expected. Can try it on another account later (or just reset my config to get the question to show up again)

PR #1744, opened 2024-09-12
Add barb training req check

Zoinkwiz:
This adds a new class, RequirementValidator, which acts as a trimmed down ConditionalStep with just requirements. The intention is for this to allow for requirements to be continually updated, in this case for barb training.

Currently there's a bit too much setup for it required I feel outside of the class itself, so I think I need to give this another look in a bit.

### DIFF
--- a/src/main/java/com/questhelper/statemanagement/BarbarianTrainingStateTracker.java
+++ b/src/main/java/com/questhelper/statemanagement/BarbarianTrainingStateTracker.java
@@ -218,12 +218,12 @@ public class BarbarianTrainingStateTracker
 			"Finished Barbarian Herblore"
 		);
 
-		reqs = new RequirementValidator(client, eventBus, (or(
+		reqs = new RequirementValidator(client, eventBus,
 			taskedWithFishing, taskedWithHarpooning, taskedWithFarming, taskedWithBowFiremaking, taskedWithPyre, taskedWithPotSmashing,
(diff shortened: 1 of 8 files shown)


src/main/java/com/questhelper/statemanagement/BarbarianTrainingStateTracker.java
```diff
@@ -0,0 +1,312 @@
...
+					"I've managed to <col=800000>smith a hasta<col=000080>!")
+			)
+		);
+
+		reqs = new RequirementValidator(client, eventBus, (or(
+			taskedWithFishing, taskedWithHarpooning, taskedWithFarming, taskedWithBowFiremaking, taskedWithPyre, taskedWithPotSmashing,
+			taskedWithSpears, taskedWithHastae, taskedWithHerblore, plantedSeed, smashedPot, litFireWithBow, sacrificedRemains, caughtBarbarianFish,
+			caughtFishWithoutHarpoon, madePotion, madeSpear, madeHasta, finishedFishing, finishedHarpoon, finishedSeedPlanting, finishedPotSmashing,
+			finishedFiremaking, finishedPyre, finishedSpear, finishedHasta, finishedHerblore
+		)));
```
pajlada, line 301: It looks weird to me that the requirements are passed in under an `or(` here. Would it be the same functionally if `RequirementValidator` instead accepted a `List<Requirement>` and we ran `List.of(...)` of all the requirements here?

pajlada (commented):
Some nits, haven't tested in-game

PR #1747, opened 2024-09-14
fix: Have wiki links try to open in browser, not local file system

Zoinkwiz:


pajlada (approved):
This was never an issue for me, but with this change everything still works as expected and is more correct. (Arch Linux btw, Firefox)

PR #1755, opened 2024-09-22
fix: Kings Random quest should have animate rock scroll as required

kcjg94:
https://oldschool.runescape.wiki/w/King%27s_Ransom

It can be obtained before the quest, so should be listed in the required items for the quest and not just the panel step.


pajlada (approved):
This is fine since the item has a tooltip mentioning it can be obtained during the quest, but the user must bring it if they have it otherwise they won't be able to get a new one. :+1:

PR #1767, opened 2024-09-26
Ethically Aquired Antiquities

Haavardaw:
This adds the new Novice quest of Ethically Aquired Antiquities, I need someone to test it! Thank you. Did a few more changes to fix up and make it functional atleast.


pajlada (changes requested):
The "start the quest" step should highlight the Civitas Illa Fortis teleport

The "Head to Port Sarim and speak to Trader Stan" step should have a dialogue highlight for "Port Sarim (wily cats)".
I don't like us suggesting two methods of teleportation. imo just suggest the Charter Ship option, and add the 3000 coins as a requirement to the "Done with museum visit" section & the above-mentioned step

"Head to Varrock Museum and speak to..." step should highlight the Varrock Teleport.
The same step should be split up into two steps
Varbit 22: "Head to Varrock Museum and speak to..."
Varbit 24: "Pickpocket Curator Haig..."

"Open the door to the room in the north east..." - since it's a "single use" door, you could highlight the door when the user is outside the orom, and highlight the crate when the user is inside the room. The text is small enough that it can stay the same for both steps

"Speak to the Curator in the centre of the Grand Museum..." should highlight the Civitas Illa Fortis teleport
It could also be renamed to "Speak to Curator Herminius in the..." to clarify which Curator you will be speaking with

 - [ ] Is the difficulty correct?
 - [ ] Are the skill requirements correct?
       missing: Quest "Shield of Arrav"
       missing: Skill "Level 25 Thieving"
 - [ ] Is the position in the optimal quest guide & optimal ironman quest guide correct?
 - [ ] Are the monsters you have to fight correctly listed?
 - [ ] Are the rewards correct?
 - [ ] Does the "RuneScape Wiki" link work?

pajlada (changes requested):
wiki link still doesn't work

PR #1777, opened 2024-09-29
feat: Remove html usage

Zoinkwiz:
HTML throws as a warning for plugin-hub PRs, so where we can it makes sense to avoid it.

Not 100% sure this has been done right, but seems pretty much right, and everything from what I've scoured looks the same with these changes.

### DIFF
--- a/src/main/java/com/questhelper/panel/QuestOverviewPanel.java
+++ b/src/main/java/com/questhelper/panel/QuestOverviewPanel.java
@@ -89,7 +89,7 @@ public class QuestOverviewPanel extends JPanel
 
 	private final JPanel introPanel = new JPanel();
 
-	private final JLabel questNameLabel = new JLabel();
+	private final JLabel questNameLabel = JGenerator.makeJLabel();
 
 	private static final ImageIcon CLOSE_ICON = Icon.CLOSE.getIcon();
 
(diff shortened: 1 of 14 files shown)


src/main/java/com/questhelper/panel/QuestOverviewPanel.java
```diff
@@ -607,11 +604,11 @@ private void updateQuestNotes(List<String> notes)
 			for (String note : notes)
 			{
 				textNote.append(note);
-				textNote.append("<br><br>");
+				textNote.append("\n\n");
```
pajlada, line 607: Unrelated to this PR, but while we're here:
Can we use a boolean to ensure we only add newlines if there are more notes?
Like this:
```java
var first = true;
for (var note : notes) {
  if (!first) {
    textNote.append("\n\n");
  }
  textNote.append(note);
  first = false;
}
```

pajlada (approved):
I have tested & confirmed the following work as expected:
 - tooltips with newlines using `\n` instead of html/br
 - `JTextArea` usage in Assist Level
 - "Skill filtering" buttons/selectors (although the button does jump in position a bit when any skills are filtered - I don't really mind)
 
image
image


 - When no quests match my filter
 - The dropdown menus for filters/difficulty/ordering
 - Herb run configuration
 - Enemies to defeat when multiple lines are listed
 - Quest Notes
 - Quest list
 - Quest selection
 - Quest steps
 - Quest sidebar category titles
 - Quest reward list

I've tested everything except actually doing a quest

pajlada (approved):

PR #1779, opened 2024-09-29
feat: Added more hints on teleports for Lost City

Zoinkwiz:


Felanbird (approved):

umer-rs:
an amulet of glory is pretty helpful for banking before going to entrana, maybe that could be added.

PR #1781, opened 2024-09-30
fix: Fall back to clickbox highlight for objects

Zoinkwiz:
This is for objects with no outline/visible aspects to highlight, such as gaps, top of stairs, etc.


pajlada (approved):
Works as expected for me
image


Without this PR applied:
image

PR #1789, opened 2024-10-03
fix: Improve logout behaviour of the assist panel

Zoinkwiz:
If you logged out after selecting a quest and  being prompted to select an assist level, it would try to direct you to a quest panel once you selected the assist level. This would occur even when you then logged out then selected it.

This ensures that the player is instead directed to the usual 'log in to see quests' page.


pajlada (approved):
Briefly tested, code looks good

Issue #1793, opened 2024-10-07
Wrong direction indicated in step


ares-tyr:
https://github.com/Zoinkwiz/quest-helper/blob/8d5bfb374b53ef59340cb9218ce4b5d968c43348/src/main/java/com/questhelper/helpers/quests/whileguthixsleeps/WhileGuthixSleeps.java#L1441

Should read:
"Stand near the map just east of the door door until Silif talks about it."
<img width="1448" alt="WGSDoorDirection" src="https://github.com/user-attachments/assets/e726577f-9274-42b9-8912-fbf51ba2629f">

PR #1794, opened 2024-10-07
Update WhileGuthixSleeps.java modify direction of Silif map step

ares-tyr:
change direction to east of door when waiting for Silif to interact with you

WGSDoorDirection

issue:  #https://github.com/Zoinkwiz/quest-helper/issues/1793


Zoinkwiz (approved):
Looks good, thanks!

PR #1807, opened 2024-10-17
fix: add hammer to dwarf cannon requirements

kcjg94:
-add item requirement for hammer
-add item requirement for hammer and rail to panel step


Zoinkwiz:
He should give you a hammer at the start of the quest if you finish the dialogue with him. I can probably make a check for if you've been given a hammer or not before guiding you to go repair though, as I imagine people see the arrow change to point elsewhere and just run off before getting it.

kcjg94:
That makes sense, I thought it was weird cause I talked to him again and he wouldn't give me a hammer.

PR #1808, opened 2024-10-18
fix: Updated widgets for Pirate's Treasure

Zoinkwiz:


pajlada (approved):
This looks good, only "issue" I could see with this is if it accidentally picks up things where it shouldn't since it does a recursive check now. Only one way to find out!

Issue #1809, opened 2024-10-18
Ordering by release date doesn't actually order by release date


Kumagor0:
Here's a screenshot of how it looks for me. The notable thing is, it is mostly correct, with some quests (like Desert Threasure II or At First Light) being obviously wrong. So maybe it's an issue with release dates of particular quests being wrong. I've also reproduced the problem on another PC and account so it's probably not just me.
![image](https://github.com/user-attachments/assets/3adadc89-0a36-45d7-9af4-e2b0e475d6e3)

Kumagor0:
Ok I think I've found where relevant code is and submitted a [PR](https://github.com/Zoinkwiz/quest-helper/pull/1810).
Full disclosure though: I've never touched any RL plugins and just edited the order in github web interface, hope it doesn't break anything.

PR #1810, opened 2024-10-18
fix: properly order all quests by release date

Kumagor0:
A possible fix for https://github.com/Zoinkwiz/quest-helper/issues/1809

### DIFF
--- a/src/main/java/com/questhelper/panel/questorders/QuestOrders.java
+++ b/src/main/java/com/questhelper/panel/questorders/QuestOrders.java
@@ -723,15 +723,19 @@ public class QuestOrders
 		QuestHelperQuest.SLEEPING_GIANTS,
 		QuestHelperQuest.THE_GARDEN_OF_DEATH,
 		QuestHelperQuest.SECRETS_OF_THE_NORTH,
+		QuestHelperQuest.DESERT_TREASURE_II,
 		QuestHelperQuest.THE_PATH_OF_GLOUPHRIE,
 		QuestHelperQuest.CHILDREN_OF_THE_SUN,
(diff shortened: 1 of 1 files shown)


pajlada:
Yours didn't compile, so I went ahead and made a new PR with this + some other changes we've wanted to make for a while

https://github.com/Zoinkwiz/quest-helper/pull/1811

Will add you as a co-author to the final squashed PR

Kumagor0:
@pajlada thank you, was it because I fixed PERILOUS_MOON to PERILOUS_MOONS without fixing [this file](https://github.com/Zoinkwiz/quest-helper/blob/0401f958ed78acce47650a698922c47b9c29ee64/src/main/java/com/questhelper/questinfo/QuestHelperQuest.java#L480) as well? Maybe you want to do that properly too if you know how to.

pajlada:
> @pajlada thank you, was it because I fixed PERILOUS_MOON to PERILOUS_MOONS without fixing [this file](https://github.com/Zoinkwiz/quest-helper/blob/0401f958ed78acce47650a698922c47b9c29ee64/src/main/java/com/questhelper/questinfo/QuestHelperQuest.java#L480) as well? Maybe you want to do that properly too if you know how to.

Yeah there's actually a few others where we don't really use the "correct" name, but I thought that'd be best to catch in a followup PR instead

`DESERT_TREASURE` should be `DESERT_TREASURE_I` now that DT2 is out
`THE_MAGE_ARENA` should be `MAGE_ARENA_I`
`THE_MAGE_ARENA_II` should be `MAGE_ARENA_II`
`ENCHANTED_KEY` should be `THE_ENCHANTED_KEY`

using the names from here: https://oldschool.runescape.wiki/w/Quests/List

PR #1812, opened 2024-10-19
fix: Update rag and bone journal check to use DIARY_TEXT

Zoinkwiz:
This resolve the issue with the the changes to the quest journal ID.

A potential issue is that this quest was already known to be a bit laggy due to the amount of checks going on, so having it check all rows for all items may make this worse.

An alternative is to wait until I'm sure on the offset of the widgets from the original positions to then keep all the checks to be a single widget check.


pajlada (approved):
I feel like with fixed offsets, it's just a matter of time before it breaks again.
I'd rather see this recursive check, which we know will always work, be optimized somehow

PR #1815, opened 2024-10-20
Turael can no longer assign sourhogs

evaan:
Small little thing I noticed when doing the quest


Felanbird (approved):
was a project rebalance skilling change

PR #1817, opened 2024-10-21
Corrected setup for cosmic rune task

### LINKED ISSUE #1813: Lumbridge Hard Diary Correction
In the Lumbridge & Draynor Hard Diaries there is a task called "Craft 56 Cosmic Runes" where the quest-helper recommends a tiara _unless essence pouches are _used.__ However, When I attempted to use essence pouches to complete the task, it did not work. I'm suggesting a simple correction to say that a tiara must be used and essence pouches will not work. Thanks for all you do!

maxwelldixon:
Updated "Tiara recommended unless using essence pouches" to "Cosmic tiara or Catalytic tiara recommended if not using abyss." Resolves #1813. 

Screenshot from 2024-10-21 12-56-57


Zoinkwiz (commented):
I'm not 100% sure, but would it make sense for the pouches to still be an option? I can't check currently, but I believe they're droppable, which means you can fill a pouch up, enter the altar with a talisman, drop the talisman and a few essence in the altar, drop the pouch, then pick up the essence so you have 28.

maxwelldixon:
@Zoinkwiz, you're correct. I just checked by entering via the cosmic altar with a filled small pouch (3), a cosmic talisman, and 26 rune essence. Then dropped the talisman, dropped 2 rune essence, emptied the small pouch (3), dropped the small pouch, and finally picked up 1 rune essence. Which filled by inventory with 28 rune essence and I crafted 56 cosmic rune simultaneously. Thank you! I didn't consider dropping 🤣

PR #1847, opened 2024-11-21
ItemRequirement rework

Zoinkwiz:
This PR intends to improve the efficiency of ItemRequirement checks by including some caching where possible. I've tried to introduce the concept of cached containers (inventory, equipped, bank), which I hope will in the future make it easier for us to include new container concepts such as the rune pouch, POH storage, etc.

I've checked how long the checks take for the sidebar each tick for the Check all items helpers as it's the largest in terms of requirements. Generally across 1000 ticks the new method while cached is 2x faster on average. I've included a 100s sample below with in the cache, with the average for it being:

```
old: 3266746 ns
new: 1408209 ns
```
[newDiff.txt](https://github.com/user-attachments/files/17851199/newDiff.txt)
[oldDiff.txt](https://github.com/user-attachments/files/17851200/oldDiff.txt)

We're not dealing with massive values in general, though I've also not encountered the lag spikes on my computer, so I'm going on the assumption for lower spec devices this will hopefully make a bigger difference.

I need to still add in some better comments and preferably some tests for this going forwards.

### DIFF
--- a/src/main/java/com/questhelper/QuestHelperPlugin.java
+++ b/src/main/java/com/questhelper/QuestHelperPlugin.java
@@ -28,6 +28,7 @@ package com.questhelper;
 import com.google.inject.Binder;
 import com.google.inject.Injector;
 import com.google.inject.Provides;
+import com.questhelper.bank.banktab.BankTabItem;
 import com.questhelper.bank.banktab.BankTabItems;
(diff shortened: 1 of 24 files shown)


pajlada (commented):
small nits

PR #1873, opened 2024-11-27
Bank improvements

Zoinkwiz:
This PR allows for the potion storage to be considered for requirements, as well as for the bank filtering for removing/adding potions from it. The bank work is based on the way the core Bank Tags plugin has done it for Layouts.

In tandem based on this, this also moves the Quest Helper filter for the bank from moving real bank items around to instead just making items into the items we're interested in, now that it's an allowed method due to implementation in the core client. This generally means quite a lot less processing per bank event.

~~Widgets which are still made (section headers and dividers, numbers for quantities needed) will now be trimmed off the size of the bank. From across my accounts the max element in BANK_ITEM_CONTAINER is always 1248, BUT this is a bit sketchy as it's possible that 1. this might change, making quest helper cut stuff off improperly, or 2. It's already wrong for some people (maybe more possible with more bank extension things?). Maybe looking at capturing the size on load before doing anything may be a better play here to ensure it's correct in more scenarios.~~
- I've updated this now to be dynamic based on initial load.

Finally, to help further with performance, I've removed the non-bank items from being part of quest filtered bank.

Across these changes I'm seeing a massive performance improvement when adding/removing items, so hopefully this is a good step in the right direction, in addition to it being good to support a new container.

Note: The quest bank can wiggle a little bit when removing items, hopefully not a big change needed to fix this but would be good to before releasing.

### DIFF
--- a/src/main/java/com/questhelper/bank/banktab/PotionStorage.java
+++ b/src/main/java/com/questhelper/bank/banktab/PotionStorage.java
@@ -34,7 +34,6 @@ import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.*;
 import net.runelite.api.events.ClientTick;
-import net.runelite.api.events.MenuOptionClicked;
 import net.runelite.api.events.VarbitChanged;
 import net.runelite.api.widgets.ComponentID;
 import net.runelite.api.widgets.Widget;
@@ -57,6 +56,7 @@ class Potion
 public class PotionStorage
 {
     static final int TOTAL_POTIONS_VARBIT = 4286;
+    static final int VIAL_IDX = 514;
     static final int COMPONENTS_PER_POTION = 5;
 
     private final Client client;
@@ -116,7 +116,7 @@ public class PotionStorage
     {
         var potionStorePotions = client.getEnum(EnumID.POTIONSTORE_POTIONS);
         var potionStoreUnfinishedPotions = client.getEnum(EnumID.POTIONSTORE_UNFINISHED_POTIONS);
-        potions = new Potion[potionStorePotions.size() + potionStoreUnfinishedPotions.size()];
+        potions = new Potion[potionStorePotions.size() + potionStoreUnfinishedPotions.size() + 1];
         int potionsIdx = 0;
         for (EnumComposition e : new EnumComposition[]{potionStorePotions, potionStoreUnfinishedPotions})
         {
@@ -142,6 +142,14 @@ public class PotionStorage
             }
         }
 
+        // Add vial
+        Potion p = new Potion();
+        p.potionEnum = null;
+        p.itemId = ItemID.VIAL;
+        p.doses = client.getVarpValue(4286);
+        p.withdrawDoses = 0;
+        potions[potions.length - 1] = p;
+
         QuestContainerManager.getPotionData().update(client.getTickCount(), getItems());
     }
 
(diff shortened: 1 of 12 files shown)


Zoinkwiz:
Main thing is I need to work out the best way to handle the scrolling when at the bottom of a section, as it always seems to jump up a bit.

src/main/java/com/questhelper/bank/banktab/QuestBankTab.java
```diff
@@ -132,14 +126,19 @@ public void shutDown()
...
-			{
-				addedWidget.setHidden(true);
-			}
-			addedWidgets.clear();
-		}
+		clientThread.invokeLater(this::removeAddedWidgets);
+	}
+	public void register(EventBus eventBus)
+	{
+		potionStorage.setQuestBankTabInterface(questBankTabInterface);
```
pajlada, line 133: With this being set in register, should it also be unset in unregister?

src/main/java/com/questhelper/bank/banktab/PotionStorage.java
```diff
@@ -0,0 +1,247 @@
...
+    static final int TOTAL_POTIONS_VARBIT = 4286;
+    static final int VIAL_IDX = 514;
+    static final int COMPONENTS_PER_POTION = 5;
+
+    private final Client client;
+    private final ItemManager itemManager;
+    private final BankSearch bankSearch;
+
+    private Potion[] potions;
+    public boolean cachePotions;
```
pajlada, line 67: Naming/documentation

If I understand it correctly, it's a hint to the PotionStorage's onClientTick function that we should look for the potion storage widget & note down what types of potions are stored in it.

src/main/java/com/questhelper/bank/banktab/PotionStorage.java
```diff
@@ -0,0 +1,247 @@
...
+    private boolean layout;
+    private Set<Integer> potionStoreVars;
```
pajlada, line 69: List of potion ItemIDs(?), or varbits(?) that are interested to us?

src/main/java/com/questhelper/bank/banktab/PotionStorage.java
```diff
@@ -0,0 +1,247 @@
...
+    static final int VIAL_IDX = 514;
+    static final int COMPONENTS_PER_POTION = 5;
+
+    private final Client client;
+    private final ItemManager itemManager;
+    private final BankSearch bankSearch;
+
+    private Potion[] potions;
+    public boolean cachePotions;
+    private boolean layout;
```
pajlada, line 68: Naming - as with the `cachePotions`, maybe an additional verb saying that "hey, this variable denotes that we'd like something to happen".

src/main/java/com/questhelper/bank/banktab/QuestBankTab.java
```diff
@@ -421,96 +611,63 @@ private int addPluginTabSection(Widget itemContainer, BankTabItems items, List<I
...
-		int totalItemsAdded = 0;
-		// Loop through all items in section
-		for (BankTabItem bankTabItem : items)
+		int count = bank.count(itemId);
+		if (count > 0)
 		{
-			boolean foundItem = false;
+			return count;
+		}
+		return potionStorage.count(itemId);
```
pajlada, line 633: Can you not store the same potion both in the bank & in the potion storage?

src/main/java/com/questhelper/requirements/item/ItemRequirement.java
```diff
@@ -540,7 +539,8 @@ private boolean checkContainersOnPlayer(Client client)
 
 	public boolean checkWithBank(Client client)
```
pajlada, line 540: Maybe rename to `checkAllContainers`?

src/main/java/com/questhelper/requirements/item/ItemRequirement.java
```diff
@@ -607,7 +611,11 @@ private ItemAndLastUpdated[] containersToCheckDefault()
 		containers.add(QuestContainerManager.getEquippedData());
 
 		if (!equip) containers.add(QuestContainerManager.getInventoryData());
-		if (shouldCheckBank) containers.add(QuestContainerManager.getBankData());
+		if (shouldCheckBank)
```
pajlada, line 614: imo the name of `containersToCheckDefault` is a bit misleading, I would personally move the logic of this function to the `check` function

src/main/java/com/questhelper/bank/banktab/PotionStorage.java
```diff
@@ -0,0 +1,247 @@
+/*
+ * Copyright (c) 2024, Adam <Adam@sigterm.info>
```
pajlada, line 2: Add yourself too - looks like you've done enough to warrant it

pajlada (commented):
Some small documentation/naming things, I have not tested this in-game (I don't have potion storage :cry: )

PR #1953, opened 2025-01-19
At First Light: Remove duplicate step

### LINKED ISSUE #1952: At First Light
The quest steps are one behind starting at the "Talk to Atza again for some trimmed fur"

vmgi:
```
talkToAtza = new NpcStep(this, NpcID.ATZA, new WorldPoint(1696, 3063, 0),
			"Talk to Atza in one of the buildings outside Civitas illa Fortis' south wall, west of the general store.",
			furSample);
		talkToAtzaAfterHandingFur = new NpcStep(this, NpcID.ATZA, new WorldPoint(1696, 3063, 0),
			"Talk to Atza in one of the buildings outside Civitas illa Fortis' south wall, west of the general store.");
		talkToAtza.addSubSteps(talkToAtzaAfterHandingFur);
```

### Problem:
'talkToAtza' is given a sub-step whose function is a duplicate. This step occurring twice makes the helper appear behind. 

### Solution:
Remove the and delete the sub-step.

After some sleuthing this is what I was able to come up with, I could have a complete misunderstanding of how things work though. 
Potentially closes #1952

### DIFF
--- a/src/main/java/com/questhelper/helpers/quests/atfirstlight/AtFirstLight.java
+++ b/src/main/java/com/questhelper/helpers/quests/atfirstlight/AtFirstLight.java
@@ -75,7 +75,7 @@ public class AtFirstLight extends BasicQuestHelper
 
 	QuestStep talkToApatura, goDownTree, talkToVerity, talkToWolf, windUpToy, useToyOnKiko,
 		checkBed, returnToWolf, goUpTree, talkToFox, takeLeaf, takeSecondLeaf, catchJerboa, useTailOnLeaves,
-		returnToFox, talkToFoxAfterPoultice, talkToAtza, talkToAtzaAfterHandingFur, makeEquipmentPile, talkToAtzaForTrim,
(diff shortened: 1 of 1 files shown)


pajlada:
I haven't done At First Light on one of my alts so I can do it in a bit and test your change alongside a polish pass, thank you!

pajlada (approved):
I tested this change and everything went smoothly for me

Could you update your PR title to match what's being changed? (e.g. `At First Light: Remove duplicate step`)

vmgi:
Sorry, just seeing this. That's great! Yes, absolutely.  :-)

Zoinkwiz (commented):
Usually how making something a substep works is that it means the substep will cause the parent step to highlight in the sidebar. In this case the substep exists as there's a point where you don't need the fur, and we don't want to incorrectly guide the player that they need more fur when they don't. 

Is the issue that the sidebar didn't correctly update to move to the `talkToAtza` step when the `talkToAtzaAfterHandingFur` step was active? Did it actively jump back to a step even earlier, and if so which one? If either of those happened I think that's an issue elsewhere, no with this here.

PR #1957, opened 2025-01-22
fix: Update container state for ItemRequirement on same tick

Zoinkwiz:
This might fix an issue where an inventory change occurs on the same tick as the check occurs.

My theory is that the item added event occurs after the inventory check for an ItemRequirement on the same tick.

This has been observed to occur for receiving items in Throne of Miscellania, and for At First Light. Ideally we can test those.


pajlada (approved):
I tested this patch while doing At First Light and did not run into any desync issues

PR #1970, opened 2025-02-07
hscb added to cbow item collection

vmgi:
### Feedback from Discord:
"also tiny thing, hscb isn't recognised as a crossbow for curse of arrav even though it is one and it can fire mith grapples"

https://oldschool.runescape.wiki/w/Hunters%27_sunlight_crossbow

### Issue: 
**Hunter's Sunlight Crossbow** is not in the **crossbows** item collection

### Implemented Solution:
add item to the collection

**The Curse of Arrav** is using `ItemCollections.CROSSBOWS` to determine whether the player has a crossbow for grappling. When investigating the item collection we see that **Hunter's Sunlight Crossbow** is not included in the collection.


pajlada:
This was already fixed in #1969 - PR-sniped

vmgi:
Ope, I was a bit hasty :-)

PR #1974, opened 2025-02-08
feat: Added Group Ironman Storage to TrackedContainers

Zoinkwiz:
This won't do anything fancy like knowing when others edit it, but you can know your last known state.

### DIFF
--- a/src/main/java/com/questhelper/panel/QuestRequirementsPanel.java
+++ b/src/main/java/com/questhelper/panel/QuestRequirementsPanel.java
@@ -296,7 +296,7 @@ public class QuestRequirementsPanel extends JPanel
 			{
 				label.setToolTipText("On steel key ring");
 			}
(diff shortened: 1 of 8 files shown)


src/main/java/com/questhelper/panel/QuestRequirementsPanel.java
```diff
@@ -297,6 +296,10 @@ else if (newColor == Color.ORANGE)
 			{
 				label.setToolTipText("On steel key ring");
 			}
+			else if (newColor == Color.YELLOW)
```
pajlada, line 299: Double check this color with https://github.com/Zoinkwiz/quest-helper/pull/1974/files#diff-3307ddf0bbb81c92f7e72c390d7b68ff6632e60154e32453f26a67ca0b877ecbR569

PR #1975, opened 2025-02-09
Add Translation Book to Item Requirements

jaybyrrd:
Noticed in the quest that the item requirement for Translation Book in Starting Off is not in the overall item requirements. Figured to add it.


pajlada (approved):
This seems fine to me since the tooltip will let you know where to get it if you threw it away

PR #1976, opened 2025-02-11
Feat: container check improvements

Zoinkwiz:
This generally looks to add a more consistent means of conveying information of which containers an item is in, and starts work towards making ItemRequirements have more defined of a state.

The sidebar now for items will always be white, or whatever you've set your 'partial success' config colour to be, if you have items generally to pass a condition, just not on you. The tooltip will let you know which containers have it.

Screenshot 2025-02-11 170749

The overlay does the same now. It may need some reconsideration in terms of default colours and readability.

Screenshot 2025-02-11 170942

Currently if you are missing some of an ItemRequirement or ItemRequirements, it will simply indicate 'failed', and not give any hints as to where what bits you do have are.

### DIFF
--- a/src/main/java/com/questhelper/QuestHelperConfig.java
+++ b/src/main/java/com/questhelper/QuestHelperConfig.java
@@ -599,6 +599,17 @@ public interface QuestHelperConfig extends Config
 	}
 
 	@ConfigItem(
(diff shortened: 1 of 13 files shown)


pajlada (approved):
Have not tested in-game, just checked the code itself

PR #1979, opened 2025-02-16
Allow volcanic mine prospector helm to be used for Falador Hard diary

coopermor:


src/main/java/com/questhelper/collections/ItemCollections.java
```diff
@@ -2184,6 +2184,11 @@ public enum ItemCollections
 		ItemID.GHOMMALS_HILT_3,
 		ItemID.GHOMMALS_HILT_2,
 		ItemID.GHOMMALS_HILT_1
+	)),
+	
+	PROSPECTOR_HELMET(ImmutableList.of(
+		ItemID.PROSPECTOR_HELMET,
+		ItemID.PROSPECTOR_HELMET_29472
```
pajlada, line 2191: [According to the wiki](https://oldschool.runescape.wiki/w/Golden_prospector_helmet), the Golden Prospector Helmet can also be used for this task. Want to add it here at the same time?

coopermor: Added, I didn't even know about this one. I just saw mine didn't work while doing the diary.

pajlada (approved):

PR #1997, opened 2025-03-22
Add Pandemonium, the first sailing quest in alpha

TTvanWillegen:
This PR adds Pandemonium, the first sailing quest in alpha.

As Sailing is alpha, there are some workarounds:
- it requires extra NpcIDs, not currently found in the RuneLite classes; these have been placed in `AlphaNpcID`.
- Sailing is not in the skill list yet; thus the experience reward is in general rewards
- Zones do not seem to work when on your boat; it might be that this is a different plane. 
- I could not find a way to check the contents of the Cargo hold; thus cannot conditionally send back to Port Sarim for Cargo.

Also, requires #1985 to build, as the old NpcIDs are no longer in the class.


Felanbird:
I'd like to give my opinion that I don't think we should add this for the Alpha, if the sailing quest is confusing then such confusion needs to be reported to Jagex, so it doesn't exist in the main version of the skill.

It defeats the purpose of Alpha feedback if we add Quest Helper arrows showing you everything to do.

Zoinkwiz:
I'm in agreement for keeping this out for the beta (is it still going actually?). Nice to have this for when it's in the main game though so thanks for the work!

PR #2007, opened 2025-04-07
fix: Allow for better handling for copied instances

Zoinkwiz:
This specifically came up in A Forgettable Tale... where the platforms with chests in the puzzle section exist a few times in the same space.

This meant sometimes the first LocalPoint found was actually the incorrect point. This change is intended to make it so that all LocalPoints can be considered by Steps before actually drawing arrows and highlights.

A note is I feel that the final section of DS2 might possibly be impacted by this, but I can't really predict in what way.

### DIFF
--- a/src/main/java/com/questhelper/steps/DetailedQuestStep.java
+++ b/src/main/java/com/questhelper/steps/DetailedQuestStep.java
@@ -295,11 +295,9 @@ public class DetailedQuestStep extends QuestStep
 			{
 				BufferedImage combatIcon = spriteManager.getSprite(location.getIconID(), 0);
 				List<LocalPoint> localPoints = QuestPerspective.getInstanceLocalPointFromReal(client, worldPoint);
-				if (localPoints.isEmpty()) continue;
 
 				for (LocalPoint localPoint : localPoints)
 				{
-					if (localPoint == null) continue;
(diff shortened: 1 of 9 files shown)


src/main/java/com/questhelper/steps/DetailedQuestStep.java
```diff
@@ -374,23 +382,24 @@ public void renderArrow(Graphics2D graphics)
 				return;
 			}
 
-			LocalPoint lp = QuestPerspective.getInstanceLocalPointFromReal(client, worldPoint);
-			if (lp == null)
-			{
-				return;
-			}
+			List<LocalPoint> localPoints = QuestPerspective.getInstanceLocalPointFromReal(client, worldPoint);
+			if (localPoints.isEmpty()) return;
```
pajlada, line 386: This check seems superfluous, unless it's faster than letting Java just not loop on line 388

pajlada (commented):
Just some questions, mostly related to nullability.
Have not tried testing this in game, but the changes _seem_ sane to me

Zoinkwiz:
Thanks @pajlada as always for the feedback! I've added in a null removal to getInstanceLocalPointFromReal, so it shouldn't contain null values within the returned list, and updated where it's used to no longer check for null nor empty, with the exception of where we're not iterating but rather taking the first element.

pajlada (approved):

PR #2009, opened 2025-04-07
fix: Allow item highlights to work for ConditionalStep

Zoinkwiz:
This should ensure that a parent step with requirements like DetailedOwnerStep and ConditionalStep have access to item highlighting. Also simplifies the divide between teleport item highlighting and normal item highlighting.


pajlada (approved):
Not tested in game, changes look reasonable

PR #2014, opened 2025-04-08
fix: Add mock for .npcs() in KeysAndLeversTest

Zoinkwiz:
I'm not very experienced with Mockito so this may be a bad way to go about this, but to fix our tests I've implemented a mock for npcs() from WorldView as that was causing issues in our KeysAndLeversTest.


pajlada (approved):
This seems like a reasonable way to do it - we might want to move the npc mocking out somewhere else if we start adding tests that rely on NPCs, but this is fine :+1:

PR #2026, opened 2025-04-09
Feat use gamevals

Zoinkwiz:


pajlada:
This looks great - can we try to make a new release, and THEN merge this in? would be great if this diff was isolated from reviewers

PR #2038, opened 2025-04-12
Herb Run Helper: Hide seed dibber when barbarian farming training is completed

coopermor:
Varbit 9613 is BRUT_MINIQUEST. My account which has completed all steps of barbarian training has a varbit value of 9. A friend who has completed all steps except for bare-handed planting has a value of 8. This leads me to believe that the least significant bit of this varbit indicates whether or not the bare-handed planting of barbarian training has been completed.


Zoinkwiz:
From my understanding that Varbit will increment for each task you do, so there's not a way to determine from it whether you've done the herb task, or all the other tasks but it for example. Only scenario you know it's done is if it's set to 11 (aka all tasks done).

I think a better way to do it would be `new RuneliteRequirement(getConfigManager(), ConfigKeys.BARBARIAN_TRAINING_FINISHED_SEED_PLANTING.getKey(), "true");`, as at least then it can be known if the player has ever opened the quest journal or done barb training with QH downloaded.

PR #2042, opened 2025-04-13
Tree run helper

Zoinkwiz:
Recreated as of all the recent changes resulting in a lot of conflicts and issues. I've fixed some shared tree area steps being skipped. Did a full run myself and seems to be working now.

Recreation of #1586

### DIFF
--- a/src/main/java/com/questhelper/helpers/mischelpers/farmruns/PaymentTracker.java
+++ b/src/main/java/com/questhelper/helpers/mischelpers/farmruns/PaymentTracker.java
@@ -24,173 +24,34 @@
  */
 package com.questhelper.helpers.mischelpers.farmruns;
 
-import com.google.common.collect.ImmutableSet;
-import java.util.Set;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import lombok.AccessLevel;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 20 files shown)


src/main/java/com/questhelper/helpers/mischelpers/farmruns/PaymentTracker.java
```diff
@@ -0,0 +1,196 @@
...
+	{
+		if (!state)
+		{
+			configManager.unsetRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, configKey(fp));
+		}
+		else
+		{
+			configManager.setRSProfileConfiguration(TimeTrackingConfig.CONFIG_GROUP, configKey(fp), state);
+		}
+	}
```
pajlada, line 167: I don't think our copy of this file should be making modifications to the time tracking config.
If we want to store our own copy of it, it should be in our own config group.

If possible, it'd be easiest if we can rely on reading TimeTrackingConfig's config value

Zoinkwiz: Good spot, I've removed any setting for it now. This means this helper won't work if the time tracking plugin isn't running, but that's probably unlikely for most people? May be worth making our own handling but really I'd want to rely on the time plugin over duplicating work.

PR #2060, opened 2025-04-17
fix: Fixed start for black arm gang

Zoinkwiz:


pajlada (approved):
Not tested, but looks good & the new IDs look good

PR #2061, opened 2025-04-17
feat: Added NER support

Zoinkwiz:
Allows user to also get item info in the NER plugin.

Preferably this can check if the player has the NER plugin downloaded before showing the option, unsure how to do that though.

### DIFF
--- a/src/main/java/com/questhelper/panel/QuestRequirementsPanel.java
+++ b/src/main/java/com/questhelper/panel/QuestRequirementsPanel.java
@@ -165,7 +165,9 @@ public class QuestRequirementsPanel extends JPanel
 
 				if (requirement instanceof ItemRequirement)
 				{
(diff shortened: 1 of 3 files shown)


src/main/java/com/questhelper/panel/QuestRequirementsPanel.java
```diff
@@ -159,6 +163,26 @@ public void actionPerformed(ActionEvent e)
 					menu.add(wikiLink);
 				}
 
+				if (requirement instanceof ItemRequirement)
+				{
+					boolean nerInstalled = questManager.getPluginManager().getPlugins().stream().anyMatch(plugin -> plugin.getName().equals("Not Enough Runes"));
+					System.out.println(nerInstalled);
```
pajlada, line 169: Get rid of this before a review gets angry :angry: 

```suggestion
```

pajlada (approved):
This seems fine - no super easy API to use to check if the plugin is enabled through PluginManager it seems

PR #2071, opened 2025-04-24
fix: Allow keyring to appear in bank for KeyringRequirement

Zoinkwiz:


src/main/java/com/questhelper/requirements/item/KeyringRequirement.java
```diff
@@ -116,6 +126,16 @@ public Color getColor(Client client, QuestHelperConfig config)
 		return this.check(client) ? config.passColour() : config.failColour();
 	}
 
+	protected String getTooltipFromEnumSet(Set<TrackedContainers> containers)
+	{
+		String basicTooltip = super.getTooltipFromEnumSet(containers);
+		if (runeliteRequirement.check())
```
pajlada, line 132: Could use the `hasKeyOnKeyRing` method you just added here to make it a bit easier to read

pajlada (approved):
Not tested in-game - this looks reasonable though

PR #2079, opened 2025-04-28
fix: Corrected order of emote steps in Varrock Medium

Zoinkwiz:


pajlada (approved):
Not tested but seems sensible

PR #2102, opened 2025-05-19
feat: Added rune pouch container tracking

Zoinkwiz:
This goes with counting the rune pouch's content as part of the inventory should it be in the inventory. If it's not, it can be counted as its own container for say the `found in X container` text.

### DIFF
--- a/src/main/java/com/questhelper/managers/QuestContainerManager.java
+++ b/src/main/java/com/questhelper/managers/QuestContainerManager.java
@@ -120,7 +120,6 @@ public class QuestContainerManager
 
         List<Item> runes = new ArrayList<>();
         final EnumComposition runepouchEnum = client.getEnum(EnumID.RUNEPOUCH_RUNE);
-        int num = 0;
 
         for (int i = 0; i < NUM_SLOTS; i++)
         {


src/main/java/com/questhelper/managers/QuestContainerManager.java
```diff
@@ -47,5 +55,72 @@ public class QuestContainerManager
...
+        QuestContainerManager.getInventoryData().update(client.getTickCount(), inventoryMap.values().toArray(new Item[0]));
+    }
+
+    static public void updateRunePouch(Client client, int varbitIdChanged)
+    {
+        if (Arrays.stream(ALL_RUNE_POUCH_VARBITS).noneMatch((varbit) -> varbit == varbitIdChanged)) return;
+
+        List<Item> runes = new ArrayList<>();
+        final EnumComposition runepouchEnum = client.getEnum(EnumID.RUNEPOUCH_RUNE);
+        int num = 0;
```
pajlada, line 108: unused
```suggestion
```

pajlada (commented):
Just a little mini PR https://github.com/Zoinkwiz/quest-helper/pull/2108 that changes the rune pouch varbit check to a set, since this will be called a lot

Tested (with my change) and it works as expected

pajlada (approved):

PR #2106, opened 2025-05-23
fix: Corrected keg of beer ID in Fremmennik Exiles

Zoinkwiz:


pajlada (approved):
ah, `VIKING_BEERKEG` is used during the fremennik trials

PR #2107, opened 2025-05-23
fix: Corrected region check for scarab dungeon in Desert Hard

Zoinkwiz:


Zoinkwiz:
Double checked on there and seems okay I reckon!

Screenshot 2025-05-23 111945
Screenshot 2025-05-23 111933

Issue #1995, opened 2025-03-13
Falador Easy Diary missing item requirement


MWTBDLTR:
Blurite Limbs task needs blurite ore/blurite bar added as a required item. I had an extra blurite in the bank for this task but the helper led me straight to the furnace in falador without requiring me to have blurite ore with me.

![Image](https://github.com/user-attachments/assets/6e586627-7ff4-497b-a4bd-e758c9a095d7)

PR #2110, opened 2025-05-27
Adds note tip to 'The Knight's Sword' to mine additional blurite

JCra-25:
Adds note tip to 'The Knight's Sword' mine additional blurite ore and to smelt an additional bar which is required for Falador Easy Diary


J-Craig97:
https://github.com/Zoinkwiz/quest-helper/issues/1995

PR #2140, opened 2025-06-28
Fix brazier directions in Enakhra's Lament

tylerdq:
The south-east and south-west directions were switched in my quest compared to the quest helper text. This change switches them back to make the helper text match the actual locations.


pajlada (approved):
Makes sense - did not test in game

Zoinkwiz:
For this, was the indicated location by the arrow (so worldpoint) also wrong, or was it just the text? There's another PR for this (https://github.com/Zoinkwiz/quest-helper/pull/2112) but they also changed the WPs I think.

tylerdq:
Unfortunately I'm not sure because I was playing on minimal assistance. In the future I'll try to switch to full assistance in situations like this before submitting. Closing this PR since there's another, more complete one pending.

PR #2141, opened 2025-06-30
Fix experience rewards list for Cold War

tylerdq:
The "Rewards:" section lists "40 Attack Experience" as a reward but this isn't listed on [the wiki](https://oldschool.runescape.wiki/w/Cold_War#Rewards) and I also didn't see that award when I completed the quest today.

This change removes the line from the rewards list.


pajlada:
This was added somewhat recently on purpose
See justification here https://github.com/Zoinkwiz/quest-helper/pull/2076

Issue #2142, opened 2025-06-30
Achievement Diary helpers missing a lot of boostable requirements


StephanBijzitter:
A lot of achievement diaries lack boost information, would you be open to me creating a pull request to add boosts in for _all_ diaries?

It's not a huge issue, but it makes using the "Show Meets Reqs" filter in the side-panel a bit useless for achievement diaries.

Closes: #1940 #1999

StephanBijzitter:
Alright, I got all the individual steps marked and the general requirements as well.

I ran into a small issue with Varrock Hard, as currently it lists 66 Hunter as requirement for mains and 69 Hunter for ironmen.
The 66 Hunter is a hard requirement for all account types that cannot be boosted, but the 69 Hunter requirement for ironmen _can_ be boosted.

However, the way the requirements were laid out it would mean that the requirement would turn orange (to indicate within boosting range) as early as 64 (+5) Hunter, which does indeed work for 69 Hunter requirement, but not for the 66 Hunter requirement.

I've changed that to simply include both, like this:
```
reqs.add(new SkillRequirement(Skill.HUNTER, 66, false));
if (questHelperPlugin.getPlayerStateManager().getAccountType().isAnyIronman())
{
		reqs.add(new SkillRequirement(Skill.HUNTER, 69, true));
}
```
Which lists both indivually for ironmen:

![Image](https://github.com/user-attachments/assets/9e18d561-d302-4c06-975e-c24054113fb0)

I feel like there may be other edge-cases similar to this, even for main accounts with the general requirements. I'll start working on figuring all of the edge-cases out and create a new pull request just for those (also giving you time to share your thoughts on how to best show these to the user), but for now I'll create the pull requests for each region.

StephanBijzitter:
Yeah right now I've only added that one edge-cases (for ironmen only) to Varrock Hard.

I think ideally it would be something like this:
Main: `66 Hunter`
Ironmen: `69 Hunter (boostable from 66)`

I haven't looked deep into that yet, but I went with the assumption that it would would be quite a big change that touches more than just diaries. Definitely up for doing something like that though

PR #2148, opened 2025-06-30
fix: add missing boost information for Karamja diaries

StephanBijzitter:
This pull request adds in boost information for every individual step of the achievement diary and includes boost information for the general requirements as well.

Relates to: #2142

### DIFF
--- a/src/main/java/com/questhelper/helpers/achievementdiaries/karamja/KaramjaHard.java
+++ b/src/main/java/com/questhelper/helpers/achievementdiaries/karamja/KaramjaHard.java
@@ -211,6 +211,7 @@ public class KaramjaHard extends ComplexStateQuestHelper
 		agility53 = new SkillRequirement(Skill.AGILITY, 53, true);
 		cooking30 = new SkillRequirement(Skill.COOKING, 30, true);
 		cooking50 = new SkillRequirement(Skill.COOKING, 50, false);
+		cooking50.setTooltip("To cook a wrapped oomlie");
 		mining52 = new SkillRequirement(Skill.MINING, 52, true);
 		ranged42 = new SkillRequirement(Skill.RANGED, 42, true);
 		runecrafting44 = new SkillRequirement(Skill.RUNECRAFT, 44, true);


src/main/java/com/questhelper/helpers/achievementdiaries/karamja/KaramjaHard.java
```diff
@@ -416,7 +424,10 @@ public List<PanelDetails> getPanels()
 		shortcutSteps.setLockingStep(usedShortcutTask);
 		allSteps.add(shortcutSteps);
 
-		PanelDetails eatOomlieWrapSteps = new PanelDetails("Eat Oomlie Wrap", Collections.singletonList(eatOomlie),
+		PanelDetails eatOomlieWrapSteps = questHelperPlugin.getPlayerStateManager().getAccountType().isAnyIronman()
+			? new PanelDetails("Eat Oomlie Wrap", Collections.singletonList(eatOomlie),
+			cooking50, oomlieWrap)
+			: new PanelDetails("Eat Oomlie Wrap", Collections.singletonList(eatOomlie),
```
pajlada, line 431: I think it would be preferred to not make any changes to this, but instead add a section-specific 50 cooking requirement for the oomlie wrap section.

Since you don't need 50 cooking for eating it, only for making it (which we don't help out with)

pajlada (changes requested):

src/main/java/com/questhelper/helpers/achievementdiaries/karamja/KaramjaHard.java
pajlada: I'm stupid - this change _is_ to the panel, not the step itself.

StephanBijzitter: Ah, yeah I was a bit confused reading this and wanted to ask for clarification :-D

StephanBijzitter:
> only feedback would be to maybe add a tooltip on the 50 cooking in the panel stating "to make the oomlie wrap"

Yeah that makes sense

PR #2186, opened 2025-07-22
fix: Allow for pure essence in ESSENCE_LOW ItemCollection

Zoinkwiz:


src/main/java/com/questhelper/collections/ItemCollections.java
```diff
@@ -1218,17 +1218,18 @@ public enum ItemCollections
...
-		ItemID.BLANKRUNE
-	)),
-
 	ESSENCE_HIGH(ImmutableList.of(
 		ItemID.BLANKRUNE_DAEYALT,
 		ItemID.BLANKRUNE_HIGH
 	)),
 
+	ESSENCE_LOW(new ImmutableList.Builder<Integer>()
+			.addAll(ESSENCE_HIGH.items).add(
```
pajlada, line 1227: This will add `BLANKRUNE_DAEYALT` again, we should remove it from ESSENCE_LOW or only add ItemID.BLANKRUNE_HIGH

PR #2190, opened 2025-07-23
feat: implement Vale Totems miniquest

pajlada:
I'd love this to be tested by someone else, since there's some randomness in the quest and I haven't covered all IDs fully.

Technically you can use magic logs or w/e, but you'd need to use the same type of shield. I have opted to use the lowest log type (Oak logs, oak shields) as the requirement, and use a tooltip to explain you can use a different type.


pajlada:
I have implemented some feedback from @coopermor.
Next person to test this, I would appreciate it if when the NPC says "...First, we'll need some lumber to rebuild the pole", that you cancel out of the conversation and see what the plugin prompts you to do.
If it's good, it should tell you to talk to the NPC still.
If it's bad, it will tell you to build the totem, which you can't until you finish the conversation.


Other thing I would appreciate someone testing: Can you try bringing something else instead of the shield? either oak longbow, oak shortbow, oak longbow (u), or oak shortbow (u)? I'd like it tested before I add it as an alternative item.

coopermor:
I tested with an oak longbow on one account and an oak shortbow on my other account and both worked for me.

PR #2192, opened 2025-07-24
fix: Allow npcs to highlight based on composition id

Zoinkwiz:


src/main/java/com/questhelper/steps/NpcStep.java
```diff
@@ -175,7 +176,7 @@ public NpcStep copy()
 	protected boolean npcPassesChecks(NPC npc)
 	{
 		if (npcName != null && (npc.getName() == null || !npc.getName().equals(npcName))) return false;
-		return npcID == npc.getId() || alternateNpcIDs.contains(npc.getId());
+		return npcID == npc.getId()  || npcID == npc.getComposition().getId() || alternateNpcIDs.contains(npc.getId());
```
pajlada: This would only allow the "main npc ID" to be compared to the NPC Composition ID, not the alternate NPC IDs. Is this intended?

Zoinkwiz: Not intended no, good spot!

PR #2194, opened 2025-07-24
feat: implement the Scrambled! quest

pajlada:
this is a draft, don't look too closely. No sidebar steps have been cleaned up, but the puzzle helper does work!


GSNotifications:
Used this solely for the puzzle and it worked perfectly fine. Thank you.

PR #2250, opened 2025-08-11
feat: Added chat command to open quest helper

Zoinkwiz:
Rough draft for a suggestion we had for allowing people to suggest helpers via chat. I think we need better matching for terms for helpers, as well as better method of looping chat to shift click boxes.

### DIFF
--- a/src/main/java/com/questhelper/QuestHelperPlugin.java
+++ b/src/main/java/com/questhelper/QuestHelperPlugin.java
@@ -46,11 +46,8 @@ import com.questhelper.util.worldmap.WorldMapAreaManager;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import net.runelite.api.*;
-import net.runelite.api.annotations.Varbit;
 import net.runelite.api.events.*;
 import net.runelite.api.gameval.InventoryID;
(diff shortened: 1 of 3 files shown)


pajlada:
This seems to work fine, I'd like to know roughly how we expect people to use this though
linking just a quest is fine (i.e. if the full message is `!qh vampyre slayer`), but trying to link it in a conversation is hit and miss, depending on if its last in a conversation or not.

Should we maybe only allow `!qh` at the start, similar to how `!pets` work?

Alternatively would be to require the quest name to be within quotation marks if it's going to use multiple words
e.g. you can type "hey gonna do !qh vampyre now" and it should only replace the `!qh vampyre` part, or "hey gonna do !qh "vampyre slayer" now" and it should only replace the `!qh "vampyre slayer"` part

diablog317:
> @diablog317 Is your preference to be able to type `!qh vampyre slayer` and it would start the quest helper for you?

yes please

diablog317:
> @diablog317 Is your preference to be able to type `!qh vampyre slayer` and it would start the quest helper for you?

Just checking in to see if this could be a possibility one day, it would be a huge help for my condition. thank you for your time

PR #2259, opened 2025-08-15
Feat reorderable herb run

Zoinkwiz:
This primarily makes the herb run match the tree run with a reorderable sidebar. This doesn't add other elements from it such as the compost checks, which would need to be added seperately.

I've adjusted the id addition for steps for sidebar ordering, so that steps within a sidebar with an id get given an id themselves, and substeps of said steps get assigned the same id. The idea is that it avoids a lot of boilerplate code potentially when it comes to reordering definitions. To ensure it works correctly I've thus added sidebar steps creation to the getSteps function, so that ids are assigned prior to steps being added to the conditional steps.

This is a bit fragile in that it might not be inherently obvious to a coder to do this currently. I think if we go with this method, I'd want to instead make all helpers make use of a global variable for sidebarSteps, make the current getPanels step basically become a `setupSidebar`, and have it as part of the things like `setupSteps`, `initializeRequirements`, etc, for all helpers.


pajlada (approved):
Works as expected for me

PR #2265, opened 2025-08-16
polish: Pirate's Treasure

pajlada:
Had to make these changes very carefully since it's one of those quests that have like 4 total states


Zoinkwiz (approved):
I believe!

PR #2277, opened 2025-08-22
Migrate sprites, objects, items, interfaces, and npcs to gamevals

coopermor:


src/main/java/com/questhelper/requirements/player/FreeInventorySlotRequirement.java
```diff
@@ -82,6 +81,6 @@ private boolean isOpenSlot(Item item)
 	@Override
 	public String getDisplayText()
 	{
-		return getNumSlotsFree() + " free " + getInventoryID().name().toLowerCase(Locale.ROOT).replaceAll("_", " ") + " slots";
+		return getNumSlotsFree() + " free inventory slots";
```
pajlada, line 84: Since `FreeInventorySlotRequirement` is only ever used for the normal inventory (which is what the class name implies too), this is fine

src/main/java/com/questhelper/steps/emote/QuestEmote.java
```diff
@@ -25,27 +25,26 @@
...
-	THINK("Think", EMOTE_THINK),
-	GOBLIN_BOW("Goblin bow", EMOTE_GOBLIN_BOW),
-	BLOW_KISS("Blow Kiss", EMOTE_BLOW_KISS),
-	IDEA("Idea", 732),
-	STAMP("Stamp", 730),
-	FLAP("Flap", 731),
-	SLAP_HEAD("Slap Head", 729),
-	SPIN("Spin", EMOTE_SPIN);
+	SKILL_CAPE("Skill Cape", SpriteID.Emotes.SKILLCAPE),
+	FLEX("Flex", SpriteID.Emotes._51),
```
pajlada, line 34: fun that it doesn't even have its own gameval definition

PR #2316, opened 2025-09-13
Add standard spellbook requirement

### LINKED ISSUE #2312: wilderness medium diary doesn't specify to be on normal spellbook despite requiring charge earth orb
Image

coopermor:
Closes #2312


pajlada (approved):
this adds it to the sidebar section, not to the diary as a whole. I think this is correct

PR #2321, opened 2025-09-14
feat: add VarbitBuilder class

pajlada:
This class is intended to help you build multiple VarbitRequirements based on a single Varbit ID

Example:
```java
var sheep1State = new VarbitBuilder(VarbitID.SHEEP_STATE);

var sheep1Prodded = sheep1State.eq(1);
var sheep1Burned = sheep1State.eq(2);
var sheep1Eaten = sheep1State.eq(3);
```

More functions for the builder similar to eq can be added easily on demand


Zoinkwiz (approved):
Neat, looks good!

PR #2358, opened 2025-10-01
fix: remove unused QuestBank references

pajlada:
In these PRs I've modified `ItemRequirement`'s `alsoCheckBank` and `useQuestBank` to longer take a `QuestBank` as a parameter as it was being stored away but never used.

I have tested these changes in-game using Bike Shedder.
At 1399, 2926, 0 we check for a regular plank without taking bank into consideration. The `Required used as conditional step passed` should only appear when the player has a plank in their inventory.
At 1399, 2925, 0 and 1399, 2924, 0 we check for a regular plank taking bank into consideration using the `alsoCheckBank` function and `setShouldcheckBank` function. The `Required used as conditional step passed` should appear either when the player has a plank in their inventory or in their bank.


src/main/java/com/questhelper/helpers/quests/beneathcursedsands/BeneathCursedSands.java
```diff
@@ -295,7 +295,7 @@ protected void setupRequirements()
 		stoneTablet.setHighlightInInventory(true);
 		chest = new ItemRequirement("Chest", ItemID.BCS_CHEST);
 		chest.setHighlightInInventory(true);
-		chest.alsoCheckBank(questBank);
+		chest.alsoCheckBank();
```
pajlada, line 298: This was wrong before since `alsoCheckBank` made a copy & modified the copy, so no modification to `chest` is actualy made here. I have purposefully not changed this behaviour as I don't know for sure which we expect to be right. I'm guessing it doesn't matter much, but since I added the lint I or someone else will notice this next time this quest gets a polish pass.

PR #2366, opened 2025-10-06
fix: One Small Favour

corderguy:
One of the first steps of the quest is to talk to a Jungle Forester south of Shilo to receive his blunt axe to be sharpened. However, the talkToJungleForestStep incorrectly listed the blunt axe as a requirement. This change removes that requirement (it's still correctly a requirement for the talkToBrian step).


pajlada (approved):
This is a bit of an old pattern of quest steps, where the step where you get a quest-specific item would have the same item listed as a requirement.
I'm fine with this change, but it's a bit subjective

corderguy:
Ah, I see. For me personally, when I was doing this quest I was confused, and went back to speak to the quest-giver in case I missed a dialog option to get the axe before moving on. I didn't keep an eye out specifically, but I don't think the same pattern is present on the following steps of the quest. It's up to y'all, of course.

PR #2396, opened 2025-10-25
Fix a couple more NPC IDs

Frosty-J:


src/main/java/com/questhelper/helpers/quests/recruitmentdrive/RecruitmentDrive.java
```diff
@@ -334,7 +334,7 @@ private QuestStep TalkToSirAmikVarze()
 
 		ObjectStep climbSecondSteps = new ObjectStep(this, ObjectID.FAI_FALADOR_CASTLE_SPIRALSTAIRS, secondStairsPosition,
 			"Climb up the stairs to talk to Sir Amik Vaze.");
-		NpcStep talkToSirAmikVarze = new NpcStep(this, NpcID.HUNDRED_VARZE, "");
+		NpcStep talkToSirAmikVarze = new NpcStep(this, NpcID.SIR_AMIK_VARZE, "");
```
pajlada, line 337: Could you instead add `SIR_AMIK_VARZE` as an alternate NPC for this step?

Frosty-J: @pajlada `hundred_varze` refers to the Recipe for Disaster version where he's sitting at the table. I don't think it's possible to start this quest from there.

PR #2400, opened 2025-10-27
Remove Cook's Helper

Zoinkwiz:
Been asked to remove it, I've left in Playermade category as it's used by the testing helper, but this should remove things directly related to the quest.


pajlada (commented):
Looks like the `QuestDetails.java` file was deleted by accident

In the `QuestHelperConfig.java` file you could technically remove the `showRuneliteObjects` entry entirely, but not important

Issue #2405, opened 2025-10-31
The Final Dawn - Various raw meats do not greenlight the item requirements


romeogiggletoes:
In the item requirements section it says you need any type of bone or raw meat, which usually this will turn green if you have the item in your inventory; however it only turns green when I have bones, not raw meat. The two I have in my inventory are specifically raw chicken and raw beef, but I'm sure there are others that don't greenlight the requirement.

<img width="1568" height="730" alt="Image" src="https://github.com/user-attachments/assets/04973c7f-5b10-47e1-8059-1146129fe4d5" />

<img width="1766" height="697" alt="Image" src="https://github.com/user-attachments/assets/89e12383-8d9b-4801-b027-c2ca006061e8" />

PR #2411, opened 2025-11-01
fix final dawn raw and cooked meat requirements

Hamish-Dickson:
Resolves https://github.com/Zoinkwiz/quest-helper/issues/2405

Also adds `ItemRequirementsHelper` which could be used in places where "Any Bones" "Any Raw Meat" is a quest requirement.
I didn't take time to replace existing usages yet, but could do if desired.


Zoinkwiz (changes requested):
Hey, looks good in-so-far as the list of items defined goes, but there's already an enum for lists of item IDs, https://github.com/Zoinkwiz/quest-helper/blob/master/src/main/java/com/questhelper/collections/ItemCollections.java. Can you add these as enums in that please so we have them all contained within the same structure, otherwise then looks good!

pajlada:
I think the order here might be a little weird - the order we add alternates in is the order we recommend the item in the bank.
It might be an idea to create one alternate with all items, ordered in the order which we might recommend the user to grab things (i.e. normal bones -> raw meat -> raw chicken -> cooked meat -> cooked chicken -> big bones -> etc...). order wouldn't have to be exact, but as long as we don't recommend superior dragon bones over raw meat

Zoinkwiz:
I think that's good, as long as we have some sensible items as the initial recommendations that can be good enough. I'd also be happy for this to have a custom order just to ensure it goes logically for both meats and bones, but that would be more work.

PR #2415, opened 2025-11-05
fix: Updated brimhaven widget highlights for tree run

Zoinkwiz:
Highlighting chartership X and the sidebar, and the spirit tree menu


pajlada (approved):
Looks good, but will have merge conflicts since you didn't pull before editing

PR #2422, opened 2025-11-09
fix: add pickaxe to overall reqs for Making Friends with My Arm

Psychemaster:


Psychemaster:
Admittedly, I hadn't reached that point in the quest when I made the PR, so I didn't even know there was a pickaxe available from the wreck - I'd actually used the quest bank tab to gather required items beforehand and it includes the pickaxe there, so wondered why it wasn't in the overall list. Now I know 😄 

There might be scope to switch the order of steps around so you search the boat before attempting to mine the cave entrance, but otherwise happy to close the PR without merging

PR #2430, opened 2025-11-19
Add Quest: Pandemonium

TTvanWillegen:
See also #1997. Requires #2435.

- Adds Pandemonium, the Sailing intro quest.

### DIFF
--- a/src/main/java/com/questhelper/helpers/quests/pandemonium/Pandemonium.java
+++ b/src/main/java/com/questhelper/helpers/quests/pandemonium/Pandemonium.java
@@ -48,6 +48,7 @@ import com.questhelper.steps.BoardShipStep;
 import com.questhelper.steps.ConditionalStep;
 import com.questhelper.steps.DetailedQuestStep;
 import com.questhelper.steps.NpcStep;
+import com.questhelper.steps.SailStep;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -260,8 +261,7 @@ public class Pandemonium extends BasicQuestHelper
 		boardShip = new BoardShipStep(this);
 		takeHelm2 = new DetailedQuestStep(this, "Navigate using the helm.");
 		raiseSails2 = new DetailedQuestStep(this, "Raise your sails.");
-		sailToPortSarim = new DetailedQuestStep(this, "Sail to Port Sarim.");
-		sailToPortSarim.setHighlightZone(portSarimDockZone);
+		sailToPortSarim = new SailStep(this, Port.PORT_SARIM);
 		letGoOfHelm = new DetailedQuestStep(this, "Let go of the helm.");
 		disembarkShipPS = new ObjectStep(this, ObjectID.SAILING_GANGPLANK_DISEMBARK, new WorldPoint(3151, 3193, 0), "Leave your ship.");
 
@@ -270,8 +270,7 @@ public class Pandemonium extends BasicQuestHelper
 		dropCargoInCargoHold = new ObjectStep(this, ObjectID.SAILING_BOAT_CARGO_HOLD_REGULAR_RAFT, "Drop the crate in your cargo hold.");
 		takeHelm3 = new DetailedQuestStep(this, "Navigate using the helm.");
 		raiseSails3 = new DetailedQuestStep(this, "Raise your sails.");
-		sailToPandemonium = new DetailedQuestStep(this, "Sail to the Pandemonium.");
-		sailToPandemonium.setHighlightZone(pandemoniumDockZone);
+		sailToPandemonium = new SailStep(this, Port.PANDEMONIUM);
 		letGoOfHelm2 = new DetailedQuestStep(this, "Let go of the helm.");
 		pickupCargoShip = new ObjectStep(this, ObjectID.SAILING_BOAT_CARGO_HOLD_REGULAR_RAFT, "Pick up the cargo from your ship.", nothingInHands);
 		boardShip3 = new BoardShipStep(this);


TTvanWillegen:
Just fixing the tests to run now.

src/main/java/com/questhelper/panel/questorders/IronmanOptimalQuestGuide.java
```diff
@@ -27,6 +27,7 @@
 
 import com.google.common.collect.ImmutableList;
 import com.questhelper.questinfo.QuestHelperQuest;
+import lombok.Getter;
```
pajlada: undo this import

.run/Application.run.xml
```diff
@@ -1,13 +0,0 @@
-<component name="ProjectRunConfigurationManager">
```
pajlada, line 1: re-add this file

src/main/java/com/questhelper/questinfo/QuestHelperQuest.java
```diff
@@ -505,6 +506,9 @@ public enum QuestHelperQuest
 	BALLOON_TRANSPORT_GRAND_TREE(new GrandTreeBalloonFlight(), "Balloon Transport - Grand Tree", QuestVarbits.BALLOON_TRANSPORT_GRAND_TREE, 1, QuestDetails.Type.MINIQUEST,
 		QuestDetails.Difficulty.MINIQUEST),
 
+	// Alpha quests
```
pajlada: can remove this comment now and move it up next to the real quests

src/main/java/com/questhelper/helpers/quests/pandemonium/Pandemonium.java
```diff
@@ -0,0 +1,321 @@
+/*
+ * Copyright (c) 2020, Zoinkwiz
```
pajlada: update the copyright to your own name & 2025

TTvanWillegen:
>Mark saw & hammer as "canBeAcquiredDuringQuest" instead of the custom tooltip, since you guide the user to get them

Done

>Make use of Gamevals in all the hardcoded object IDs

Done

>For the first step,specify that Will & Anne are on the docks of Port Sarim, also add puncutation. Also add a dialog
highlight for "Yes." instead of "Yes".

Done

>You can make your first two steps "Talk to Will or Anne..." and "Finish your interview with Will and Anne" into just one
step instead. So just remove completeInterview and use talkToWill both for 0 and 2 quest states.

Done

>For "Board the ship" - clarify that you need to talk to Will or Anne to board the ship. Don't use a regex matcher, the
options are "I guess I'm ready..." and "I guess so...". Have you seen the "Okay" option somewhere?

Done, you get the "Okay" option if you have teleported away from the ship and have to get back.

>"Listen to the job explanation" should be an NpcStep on Will & Anne, with the NPC IDs being 14964 and 14959.

... It is? Highlighting doesn't work on ships.

>The helper jumps straight from "Listen to the job explanation" to "Learn about the other part of your job". The order of
"addStep" in a ConditionalStep matter - so you need to move the navigateShip etc steps above explainJob2. Additionally,
you can use the and function instead of using new Conditions(LogicType.AND).
I wonder if we should just have a step that says "Follow the in-game guide to the Salvaging wreck"? The client already
provides arrows.
We could provide an arrow for where they need to go once the player is already sailing.

Done. I wrongly had `notAtShipwreck` instead of `atShipwreck`. 
We can still change to "Follow ingame guide" if preferred? Does that still work if you teleport away from the ship?

>The "Deploy the Salvaging Hook" step is activated too early - quest step needs to be set to 14 before they can salvage.

Fixed.

>"Learn where you are." should be worded something like "Talk to 'Squawking' Steve Beanie behind the bar to learn about
where you are."
"Learn who Will and Anne area." should be worded something like "Talk to 'Squawking' Steve Beanie about Will and Anne."

Done

>When talking to Ribs, technically both options are valid - maybe highlight both?

Done.

>"Ask Steve about Will and Anne's location" - again clarify where Steve is. A lot of players got lost here

Done.

>"Ask Jim about getting a boat" - use the full name "Junior Jim" and a small description for where he is, maybe just "by
the docks" is fine

Done

>"Pick up a Saw" - add "from the crate of saws"
>"Pick up a Hammer" - add "from the crate of hammers"

Done

>"Disembark your vessel" isn't highlighting the object to disembark - not sure if this is solvable easily so feel free to
leave this.

As we discussed; highlighting on ships doesn't work yet and warrants a different PR.

>You don't need to Dock at the Port Sarim bouy

Removed.

>The conditional steps under the "You got the job... again!" for sailing to Port sarim are in the wrong order, so you
never get told to head onto the port sarim docks. I think the "disembarkShipPS" step needs to be added to
cSailToPortSarim too

Ah, I was testing things here and forgot to change them back. where it said `sailToPortSarim` in the startstep of `cCheckPortmaster`, it should've been `cSailToPortSarim`. That should make it work.

>"Pick up the cargo from the table" should have a "noithing equipped in your main hand or offhand" requirement

Added.

>Deliver the cargo:
>No need to dock at the Pandemonium buoy, just let go of the helm

Removed.

>When arriving at the Pandemonium docks, you're prompted to sail to Port Sarim again. Something with the conditional
steps here is wrong. I wasn't told to pick up my cargo or leave the vessel

I'd have to check this again.

>"Deliver the cargo at the table" - clarify that it's the ledger table

Done

>The "Meet Grog at the bar" highlights Steve too, maybe just highlight Grog instead even though both options work.

Changed to "Meet 'Squawking' Steve Beanie and Old Grog at the bar."

PR #2431, opened 2025-11-20
Add charting helper

Zoinkwiz:
This is a helper intended to make it less annoying to follow the list of things that need charting. I've used the wiki and the cache to determine most of this.

Currently a few things of note which make this not fully complete:
1. Some WorldPoints still missing (need to add once community finds things)
2. ~~No definition of NpcID or ObjectID yet. Using a generic pointer to where they should be for now. Need to go through at some point and add proper highlighting via ID. Ideally we'd find a nice way to do this not manually due to how many there are, but if not will just need to be a bit of a slog.~~ **Added handlers for each case.**
3. I've got attribution to the wiki in the generated file of details, but I'd like to add further attribution to the player-facing side as well.
4. ~~Still need to decide what is best for steps you don't have the level or quest req for. Perhaps we hide and skip them until you have the req. A newer possible feature idea would be to grey out and have it skipped over if missing the req for it.~~ **added a faded effect**
5. Ordering is not optimal for various sea

### DIFF
--- a/src/main/java/com/questhelper/helpers/activities/charting/ChartingTasksData.java
+++ b/src/main/java/com/questhelper/helpers/activities/charting/ChartingTasksData.java
@@ -219,8 +219,7 @@ public final class ChartingTasksData
 			new ChartingTaskDefinition(ChartingType.GENERIC, "Find the wreck of the Counterweight north west of Deepfin Point.", new WorldPoint(1845, 2824, 0), "Shrouded Ocean", 1, VarbitID.SAILING_CHARTING_GENERIC_WRECK_WEIGHT_COMPLETE),
 			new ChartingTaskDefinition(ChartingType.SPYGLASS, "Use your spyglass to get a good view of the south eastern temple in the Avium Savannah from the south.", new WorldPoint(1737, 2907, 0), "Shrouded Ocean", 1, VarbitID.SAILING_CHARTING_SPYGLASS_AUREUM_COAST_COMPLETE),
 			new ChartingTaskDefinition(ChartingType.CRATE, "Find a Sealed crate in the middle of the Aureum Coast and sample the contents.", new WorldPoint(1762, 2854, 0), "Shrouded Ocean", 12, VarbitID.SAILING_CHARTING_DRINK_CRATE_PORTAL_PERRY_COMPLETE),
-			new ChartingTaskDefinition(ChartingType.DIVING, "With help from a mermaid guide, document the water depth south of Vatrachos Island.", new WorldPoint(1889, 2882, 0), "Shrouded Ocean", 38, VarbitID.SAILING_CHARTING_MERMAID_GUIDE_AUREUM_COAST_COMPLETE,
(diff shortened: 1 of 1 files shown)


src/main/java/com/questhelper/helpers/activities/charting/ChartingTasksData.java
```diff
@@ -0,0 +1,549 @@
...
+public final class ChartingTasksData
+{
+	private ChartingTasksData() {}
+
+	private static final List<ChartingSeaSection> SECTIONS = List.of(
+		new ChartingSeaSection("Kharidian Sea", List.of(
+			new ChartingTaskDefinition("Generic", "Find a crashed glider south of the Karamja Shipyard.", new WorldPoint(2987, 3012, 0), "Ardent Ocean", 1, VarbitID.SAILING_CHARTING_GENERIC_GLIDER_KHARIDIAN_SEA_COMPLETE),
+			new ChartingTaskDefinition("Generic", "Find the wreck of the Pandemonium II west of the Pandemonium.", new WorldPoint(2973, 2996, 0), "Ardent Ocean", 1, VarbitID.SAILING_CHARTING_GENERIC_PANDEMONIUM_2_COMPLETE),
+			new ChartingTaskDefinition("Generic", "Find the wreck of the Pandemonium III west of the Kharidian Bandit Camp.", new WorldPoint(3143, 2981, 0), "Ardent Ocean", 1, VarbitID.SAILING_CHARTING_GENERIC_PANDEMONIUM_3_COMPLETE),
+			new ChartingTaskDefinition("Spyglass", "Use your spyglass to admire Chartin' Charles McAtless in the cave on the Pandemonium.", new WorldPoint(3051, 9390, 0), "Ardent Ocean", 1, VarbitID.SAILING_CHARTING_SPYGLASS_CHARTING_TUTOR_COMPLETE),
```
pajlada:
```suggestion
			new ChartingTaskDefinition("Spyglass", "Use your spyglass to admire Chartin' Charles McAtless in the cave on the Pandemonium.", new WorldPoint(3051, 9389, 0), "Ardent Ocean", 1, VarbitID.SAILING_CHARTING_SPYGLASS_CHARTING_TUTOR_COMPLETE),
```

src/main/java/com/questhelper/helpers/activities/charting/ChartingTasksData.java
```diff
@@ -0,0 +1,549 @@
...
+			new ChartingTaskDefinition("Spyglass", "Use your spyglass to get a good view of the Pandemonium from the cave entrance on the island.", new WorldPoint(3045, 2997, 0), "Ardent Ocean", 1, VarbitID.SAILING_CHARTING_SPYGLASS_THE_PANDEMONIUM_COMPLETE)
```
pajlada:
```suggestion
			new ChartingTaskDefinition("Spyglass", "Use your spyglass to get a good view of the Pandemonium from the cave entrance on the island.", new WorldPoint(3045, 2996, 0), "Ardent Ocean", 1, VarbitID.SAILING_CHARTING_SPYGLASS_THE_PANDEMONIUM_COMPLETE)
```

Zoinkwiz:
I think major things now still needed are:

1. Attach the answer for each mermaid to the appropriate step
2. Ordering in a way that makes sense for a player making their way through the skill

If those are done and we get the final coords made available over time, I think this will be in a usable state.

PR #2434, opened 2025-11-21
[feat-charting-helper] Implement weather steps

claudiodekker:
Coordinates are from the wiki: https://oldschool.runescape.wiki/w/Meaty_Aura_Logist?action=edit&veswitched=1
The test failure are pre-existing from the source branch.

https://github.com/user-attachments/assets/2e985b4f-9854-4ea8-8d31-29d5ce629b6f


Zoinkwiz (approved):
Can't test the locations but tested logic with some fake other items and works as expected.

PR #2438, opened 2025-11-23
Skiff for Drumstick Isle

Frosty-J:
Crystal-flecked waters require an adamant keel (66 Sailing and 62 Construction) or better to navigate. Probably boostable?

Image

Image


Zoinkwiz:
Just tried this myself, was able to get it by just staying out of the crystal water and getting it from a distance.

Screenshot 2025-11-29 at 23 42 46

Issue #2342, opened 2025-09-21
Land of the Goblins: Quest highlighting bug


CloudS3c:
- [ ] Dialogue highlighting missing for High Priest Bighead (True/False/I Don’t Know)
- [ ] Aggie dialogue option was not highlighted too the (Can you make dyes for me please?)

PR #2442, opened 2025-11-23
fix: make "last line condition" dialog step more explicit

TTvanWillegen:
Resolves (part of) #2342


src/main/java/com/questhelper/steps/QuestStep.java
```diff
@@ -370,8 +370,8 @@ public QuestStep addDialogSteps(String... newChoices)
 
 	public QuestStep addDialogConsideringLastLineCondition(String dialogString, String choiceValue)
```
pajlada: Could you rename the second param at the same time?
from `choiceValue` to `expectedPreviousLine`

pajlada: or first param i guess is actually correct

pajlada (approved):

TTvanWillegen:
Like said in discord; LGTM!

I prefer it over the long method, especially the one in the Current Affairs PR.

PR #2448, opened 2025-11-24
Sailing coord fixes

Zoinkwiz:
Creates DefinedPoint, which is intended to make it clearer when we're using a WorldPoint which needs translation before usage or not.

Makes use of worldviews for the player and the toplevel across usual wp usage.

### DIFF
--- a/src/test/java/com/questhelper/MockedTest.java
+++ b/src/test/java/com/questhelper/MockedTest.java
@@ -53,6 +53,7 @@ import org.mockito.Mockito;
 import javax.inject.Named;
 import java.awt.image.BufferedImage;
 import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.ScheduledExecutorService;
 
 import static org.mockito.Mockito.*;
@@ -141,7 +142,8 @@ public abstract class MockedTest extends MockedTestBase
 
 		WorldView mockedWorldView = mock(WorldView.class);
 		when(mockedWorldView.getScene()).thenReturn(mockedScene);
-		when(mockedWorldView.getPlane()).thenReturn(-1);
+		when(mockedWorldView.getPlane()).thenReturn(0);
+		when(mockedWorldView.isTopLevel()).thenReturn(true);
 
 		@SuppressWarnings("unchecked")
 		IndexedObjectSet<? extends NPC> npcSetMock = (IndexedObjectSet<? extends NPC>) mock(IndexedObjectSet.class);
@@ -153,6 +155,10 @@ public abstract class MockedTest extends MockedTestBase
 		var mockedItemContainer = Mockito.mock(ItemContainer.class);
 		when(mockedItemContainer.getItems()).thenReturn(new Item[0]);
 		when(client.getItemContainer(anyInt())).thenReturn(mockedItemContainer);
+
+		IndexedObjectSet<? extends WorldView> wvSetMock = (IndexedObjectSet<? extends WorldView>) mock(IndexedObjectSet.class);
+		when(wvSetMock.iterator()).thenReturn(Collections.emptyIterator());
+		doReturn(wvSetMock).when(mockedWorldView).worldViews();
 
 		var mockedPlayer = Mockito.mock(Player.class);
 		when(mockedPlayer.getLocalLocation()).thenReturn(new LocalPoint(1, 1, 1));


src/main/java/com/questhelper/steps/tools/DefinedPoint.java
```diff
@@ -0,0 +1,126 @@
...
+ * or rendering should go through {@link QuestPerspective} helpers so WorldView translations are handled
+ * consistently.
+ */
+@Getter
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public final class DefinedPoint
+{
+	private final WorldPoint worldPoint;
+
+	public static DefinedPoint of(WorldPoint worldPoint)
```
pajlada: can you tag both the return value & the arg as `@Nullable`?

src/main/java/com/questhelper/requirements/RegionHintArrowRequirement.java
```diff
@@ -1,64 +0,0 @@
...
- * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
- */
-package com.questhelper.requirements;
-
-import com.questhelper.requirements.zone.Zone;
-import com.questhelper.steps.tools.QuestPerspective;
-import net.runelite.api.Client;
-import net.runelite.api.coords.WorldPoint;
-
-public class RegionHintArrowRequirement extends SimpleRequirement
```
pajlada, line 32: random comment for myself: This requirement was unused, that's why it's removed

src/main/java/com/questhelper/requirements/location/TileIsLoadedRequirement.java
```diff
@@ -49,14 +49,21 @@ public class TileIsLoadedRequirement extends AbstractRequirement
 	public TileIsLoadedRequirement(WorldPoint worldPoint)
 	{
 		assert(worldPoint != null);
-		this.worldPoint = worldPoint;
-		this.displayText = "WorldPoint " + worldPoint.toString() + "is loaded locally.";
+		definedPoint = DefinedPoint.of(worldPoint);
+		this.displayText = "WorldPoint " + definedPoint.getWorldPoint().toString() + "is loaded locally.";
```
pajlada: this could call `this(DefinedPoint.of(worldPoint);` instead

src/main/java/com/questhelper/steps/tools/QuestPerspective.java
```diff
@@ -333,4 +357,158 @@ private static void addToPoly(Client client, Polygon areaPoly, WorldPoint wp, in
...
 		}
 	}
+
+	/**
+	 * Compares a quest-defined {@link WorldPoint} with a runtime {@link LocalPoint}, normalizing
+	 * both into the top-level world space so they can be compared safely across WorldViews.
+	 *
+	 * @return {@code true} when both coordinates refer to the same tile in the main world.
+	 */
+	public static boolean matchesWorldPoint(Client client, WorldPoint definedWorldPoint, LocalPoint runtimeLocalPoint)
```
pajlada: Should this be a `DefinedPoint definedPoint`? Seems like it and the other overload only ever get called from within `DefinedPoint`.

I think that would make sense if the intent for `DefinedPoint` is to make misuse of coordinates more difficult

src/main/java/com/questhelper/steps/QuestStep.java
```diff
@@ -636,8 +637,7 @@ protected void renderInventory(Graphics2D graphics, WorldPoint worldPoint, List<
 				if (distanceLimit)
 				{
 					WorldPoint playerLocation = client.getLocalPlayer().getWorldLocation();
-					WorldPoint goalWp = QuestPerspective.getWorldPointConsideringWorldView(client, client.getTopLevelWorldView(), worldPoint);
-					if (goalWp != null && playerLocation.distanceTo(goalWp) <= 100) continue;
+					if (definedPoint.distanceTo(playerLocation) <= 100) continue;
```
pajlada: It looks like `QuestPerspective.getWorldPointConsideringWorldView` used to handle `worldPoint` being null, but we don't do a `definedPoint` null check here.
The null check should probably be added back, unless you've ensured the paths to this function already have a check, in which case we should mark the `DefinedPoint` as `@Nonnull`

src/main/java/com/questhelper/helpers/quests/lunardiplomacy/MemoryChallenge.java
```diff
@@ -105,15 +105,11 @@ public void onGameTick(GameTick event)
 
 	public void checkNextTile(int wpsPos)
 	{
-		WorldPoint instanceWp = QuestPerspective.getWorldPointConsideringWorldView(client, client.getTopLevelWorldView(), wps.get(wpsPos));
-		if (instanceWp == null)
-		{
-			return;
-		}
+		var definedPoint = DefinedPoint.of(wps.get(wpsPos));
```
pajlada, line 108: Do we need to check if `definedPoint` is null here?

src/main/java/com/questhelper/steps/ConditionalStep.java
```diff
@@ -422,12 +423,12 @@ public void makeWidgetOverlayHint(Graphics2D graphics, QuestHelperPlugin plugin)
 		{
 			currentStep.makeWidgetOverlayHint(graphics, plugin);
 		}
-		WorldPoint activeWp = (currentStep instanceof DetailedQuestStep) ? ((DetailedQuestStep) currentStep).getWorldPoint() : null;
+		DefinedPoint activeDp = (currentStep instanceof DetailedQuestStep) ? ((DetailedQuestStep) currentStep).definedPoint : null;
```
TTvanWillegen:
```suggestion
		DefinedPoint activeDp = (currentStep instanceof DetailedQuestStep) ? ((DetailedQuestStep) currentStep).getDefinedPoint(): null;
```

src/main/java/com/questhelper/steps/DetailedQuestStep.java
```diff
@@ -247,7 +248,7 @@ public void onGameStateChanged(final GameStateChanged event)
 
 	public void setWorldPoint(WorldPoint worldPoint)
```
TTvanWillegen: I feel like there is also going to be a need for 
```java
public void setWorldPoint(DefinedPoint definedPoint)
```
Maybe move logic there and just make this method the wrapper.

src/main/java/com/questhelper/steps/ObjectStep.java
```diff
@@ -100,7 +101,7 @@ public ObjectStep(QuestHelper questHelper, int objectID, WorldPoint worldPoint,
 
 	public ObjectStep copy()
 	{
-		ObjectStep newStep = new ObjectStep(getQuestHelper(), objectID, worldPoint, null, requirements, recommended);
+		ObjectStep newStep = new ObjectStep(getQuestHelper(), objectID, definedPoint.getWorldPoint(), null, requirements, recommended);
```
TTvanWillegen: With all these; how clean is it to unwrap and then immediately re-wrap the points?

src/main/java/com/questhelper/steps/QuestStep.java
TTvanWillegen: DefinedPoint has 
```java
public static DefinedPoint of(WorldPoint worldPoint)
	{
		return worldPoint == null ? null : new DefinedPoint(worldPoint);
	}
```
so it is clearly nullable, requiring the null check here.

pajlada: The reason I'm not saying we 100% need it is because of the "unless you've ensured the paths to this function already have a check, in which case we should mark the `DefinedPoint` as `@Nonnull`" statement

PR #2449, opened 2025-11-25
Feat: Adds the locations for all the other ports.

TTvanWillegen:
Adds the locations for all the other ports.
Locations should be as below:
image


TTvanWillegen:
@Zoinkwiz Prif has been added, so this PR should add all the main ports (not the dock points)

PR #2463, opened 2025-11-27
feat: Added boat state tracking

Zoinkwiz:
Allows for info about boat state, both one you're on and boats in your collection, to be persisted and used for requirements.

### DIFF
--- a/src/main/java/com/questhelper/helpers/activities/charting/ChartingTasksData.java
+++ b/src/main/java/com/questhelper/helpers/activities/charting/ChartingTasksData.java
@@ -48,14 +48,13 @@ public final class ChartingTasksData
 			new ChartingTaskDefinition(ChartingType.CURRENT, "Test the currents south west of the Kharidian Bandit Camp.", new WorldPoint(3145, 2963, 0), "", 22, VarbitID.SAILING_CHARTING_CURRENT_DUCK_KHARIDIAN_SEA_COMPLETE)
 			)),
 		new ChartingSeaSection(2, "Lumbridge Basin", List.of(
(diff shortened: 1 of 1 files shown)


src/main/java/com/questhelper/requirements/sailing/BoatResistanceType.java
```diff
@@ -0,0 +1,260 @@
...
+	CRYSTAL("crystal", VarbitID.SAILING_SIDEPANEL_BOAT_CRYSTALFLECKED_RESISTANT, false),
+	FETID_WATER("fetid water", VarbitID.SAILING_SIDEPANEL_BOAT_FETIDWATER_RESISTANT, false),
+	RAPID("rapid", VarbitID.SAILING_SIDEPANEL_BOAT_RAPIDRESISTANCE, true),
+	STORM("storm", VarbitID.SAILING_SIDEPANEL_BOAT_STORMRESISTANCE, true);
+
+	private final String displayName;
+	private final int varbitId;
+	private final boolean showLevel;
+
+	public int getActiveBoatLevel(Client client)
```
pajlada, line 47: `getActiveBoatLevel` returns the current boat's resistance level to the given type, while `getBoatSlotLevel` returns the boat in the given slot's resistance level to the given type right?

pajlada (approved):

src/main/java/com/questhelper/requirements/sailing/BoatResistanceType.java
Zoinkwiz: That's what it should do from my thinking on it, yeah. The thinking being we can show if the player is currently on a boat which has the right resistances to do something (which can also be another player's boat), or we can indicate if a player has the requisite components on their boat for doing something prior to getting onto a boat.

PR #2496, opened 2025-12-17
Fix skill requirement being set to boostable

jespernapp:


pajlada (approved):
The general requirement was changed to be marked as unboostable a few months ago, but this panel requirement was missed.
Thanks!

PR #2499, opened 2025-12-19
Tree run decoupling

maskarauder:
Inner panels such as the two within the Farming Guild section are not reorganizable.
Additionally, it wants to select the Fruit Tree first despite appearing 2nd in the list.


maskarauder:
Sorry for the Git incompetence, didn't realize force pushing to my own project would propagate across and cause issues with the PR.

PR #2540, opened 2026-01-31
fix: Vale Totems does not continue after building a non-oak base

### LINKED ISSUE #2388: Vale Totems does not continue after building a non-oak base
The check for the totem base varbit checks explicitly for 1.
Magic logs make that varbit 5 instead.
Solution is to check >0 in the `isTotemBaseBuilt` varbit requirement

ZeynelTok:
should fix #2388 as per requirement.

There is also an unused import in `ValeTotems.java` that I can clean up if required.


pajlada (approved):
This makes sense to me. Thanks for noticing _and_ for fixing it!

Feel free to remove the unused import if you'd like, but it's not a requirement

PR #2576, opened 2026-02-15
fix tiara requirement name

### LINKED ISSUE #2555: Falador Easy Diary item name error
Not sure if its a typo but the item silver tiara does not exist. It's just called Tiara.

Image
Image

euclio:
Fixes #2555.

Instead of just fixing the typo, I introduced a new factory for constructing `ItemRequirement`s. Now helpers can construct `ItemRequirement`s with just an item ID and the plugin will default to using the in-game name for the item, instead of needing to supply a redundant item name. This will make issues like this less likely going forward.

Although the factory is quite small, I can imagine adding similarly helpful construction functions to it in the future for other kinds of duplication (such as common requirements like "Any pickaxe").


pajlada:
I'm not really fond of this solution, and would rather see just the typo fixed.

PR #2589, opened 2026-02-22
polish: Vampyre Slayer

pajlada:
I made DialogRequirement able to read MESBOX entries. There's a MesboxRequirement but I specifically want to be able to read both things so it can be unmarked as active when a normal DIALOG message appears.
Could maybe be generalized, but this seemed like the simplest solution for now. I made the flag if-statement kind of verbose because any attempt at minimizing it made it impossible to reason about.

Quest-related fixes:
 - Fixes the quest start dialog highlight
 - Made the garlic acquisition optional, with some instructions to the player for how to tick off the section to skip it
 - Guide the player into buying beer - this is what added the most complexity. The beer is taken away, but no item is given and no varbit is changed so the only way to figure out that you're still in that dialog is to check on your DIALOG and MESBOX messages, hence the unrelated change above.
 - Marked the garlic as recommended instead of required in the sections/steps.


src/main/java/com/questhelper/requirements/npc/DialogRequirement.java
```diff
@@ -40,10 +40,16 @@ public class DialogRequirement extends SimpleRequirement
 	@Setter
 	String talkerName;
 	final List<String> text = new ArrayList<>();
-	final boolean mustBeActive;
+
+	@Setter
+	boolean mustBeActive;
```
Zoinkwiz, line 46: Currently I think all other usage of setting this via the constructor set it to false. I think in my mind we want this to consider effectively if there is an active dialog box with the expected dialog on the screen at any time, which means say if we close the dialog box then the requirement fails.

Currently we only do this fail check when a new dialog loads. I've made a PR (https://github.com/pajlada/quest-helper/pull/8) which should hopefully also allow this to work if you agree with the approach. We could also just every tick check if a dialog box exists and has the expected text, but I think it's nice to avoid as much as possible things which have to run every tick.

pajlada: Awesome, will give this change a try

pajlada: This worked as expected for me :+1: Thanks!

PR #2595, opened 2026-02-27
polish(Monk's Friend): add tooltip to jug of water

cbrintnall:
Just a small change, since the nearby shops apparently don't sell jugs. Adds a tooltip that points out the nearest jug (filled with water) you can find.

### DIFF
--- a/src/main/java/com/questhelper/helpers/quests/monksfriend/MonksFriend.java
+++ b/src/main/java/com/questhelper/helpers/quests/monksfriend/MonksFriend.java
@@ -91,7 +91,7 @@ public class MonksFriend extends BasicQuestHelper
 
 		log = new ItemRequirement("Logs", ItemID.LOGS);
 		jugOfWater = new ItemRequirement("Jug of Water", ItemID.JUG_WATER);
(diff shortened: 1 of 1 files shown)


src/main/java/com/questhelper/helpers/quests/monksfriend/MonksFriend.java
```diff
@@ -91,6 +91,7 @@ protected void setupRequirements()
 
 		log = new ItemRequirement("Logs", ItemID.LOGS);
 		jugOfWater = new ItemRequirement("Jug of Water", ItemID.JUG_WATER);
+		jugOfWater.setTooltip("You can find one in the Yanille cook shop");
```
pajlada: Labeled as a "cooking shop" on the wiki https://oldschool.runescape.wiki/w/Yanille#Cooking_Shop


```suggestion
		jugOfWater.setTooltip("You can find one in the Yanille cooking shop");
```

odderb: Fixed, thanks.

pajlada (approved):

PR #2604, opened 2026-03-04
Mage arena 2 improvements

Zoinkwiz:
This primarily does 3 things:

1. Adds the accurate position for the second spot near the rogue's castle
2. Adds a region highlight for each spot to indicate where you can stand
3. Adds handling for detecting mouse clicks on the god names for scanning, as well as key press detection. This means we can ensure that we know what god the hot/cold text appearing is relavant for.

### DIFF
--- a/src/main/java/com/questhelper/helpers/miniquests/themagearenaii/MageArenaBossStep.java
+++ b/src/main/java/com/questhelper/helpers/miniquests/themagearenaii/MageArenaBossStep.java
@@ -332,12 +332,6 @@ public class MageArenaBossStep extends DetailedQuestStep
 	}
 
 	@Override
-	public void onGameTick(GameTick gameTick)
-	{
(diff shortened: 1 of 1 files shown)


src/main/java/com/questhelper/helpers/miniquests/themagearenaii/MageArenaBossStep.java
```diff
@@ -201,9 +259,71 @@ public void makeWorldOverlayHint(Graphics2D graphics, QuestHelperPlugin plugin)
...
+		{
+			lastGodClicked = godClicked;
+		}
+	}
+
+	@Override
+	public void onGameTick(GameTick gameTick)
+	{
+		super.onGameTick(gameTick);
+	}
```
pajlada: This override should not be necessary

PR #2622, opened 2026-03-05
Update Lumbridge diary elite step and add logic for using fairy ring

### LINKED ISSUE #2620: Lumbridge Elite Diary Step Replaced
The Lumbridge & Draynor Diary Elite pickpocket step was replaced with a mithril grapple step. 

I was a little confused when doing the diary that it auto-completed after doing the grapple, before trying to pickpocket the NPC. I checked the Wiki and noticed the step was updated, but not Quest Helper. 

[10 July 2024: The task of pickpocketing Movario has been replaced with grappling a pylon on the Dorgesh-Kaan Agility Course.
](https://oldschool.runescape.wiki/w/Lumbridge_%26_Draynor_Diary#Changes)

henrywild:
Resolves https://github.com/Zoinkwiz/quest-helper/issues/2620.

Also slightly expands Dorgesh-Kaan agility course zone to the north to fix steps reverted near stairs to 2nd floor of city as can be seen in screenshots.
LumbridgeEliteGrappleTaskWithinAgiCourseZone
LumbridgeEliteGrappleTaskWrongAgiCourseZone

### DIFF
--- a/src/main/java/com/questhelper/helpers/achievementdiaries/lumbridgeanddraynor/LumbridgeElite.java
+++ b/src/main/java/com/questhelper/helpers/achievementdiaries/lumbridgeanddraynor/LumbridgeElite.java
@@ -65,23 +65,23 @@ public class LumbridgeElite extends ComplexStateQuestHelper
 	// Items required
 	ItemRequirement lockpick, crossbow, mithgrap, lightsource, axe, addyBar, hammer, essence, waterAccessOrAbyss, qcCape;
 
-	// Items recommended
-	ItemRequirement ringOfDueling, dorgSphere;
+	// Recommended teleport items
+	ItemRequirement ringOfDueling, dorgSphere, dramenStaff;
 
-	Requirement notRichChest, notMovario, notChopMagic, notAddyPlatebody, notWaterRunes, notQCEmote, allQuests,
+	Requirement notRichChest, notGrapplePylon, notChopMagic, notAddyPlatebody, notWaterRunes, notQCEmote, allQuests,
 		deathToDorg, templeOfIkov;
 
-	QuestStep claimReward, richChest, movario, chopMagic, addyPlatebody, waterRunes, qcEmote, moveToWater,
-		dorgStairsChest, dorgStairsMovario, moveToOldman, moveToUndergroundChest,
-		moveToUndergroundMovario, moveToDorgAgi;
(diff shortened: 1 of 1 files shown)


src/main/java/com/questhelper/helpers/achievementdiaries/lumbridgeanddraynor/LumbridgeElite.java
```diff
@@ -192,29 +199,32 @@ public void setupSteps()
...
-		moveToDorgMovario.addAlternateObjects(ObjectID.CAVE_GOBLIN_CITY_DOORL);
+		moveToDorgPylon.addAlternateObjects(ObjectID.CAVE_GOBLIN_CITY_DOORL);
 
-		dorgStairsMovario = new ObjectStep(this, ObjectID.DORGESH_1STAIRS_POSH, new WorldPoint(2721, 5360, 0),
+		dorgStairsPylon = new ObjectStep(this, ObjectID.DORGESH_1STAIRS_POSH, new WorldPoint(2721, 5360, 0),
 			"Climb the stairs to the second level of Dorgesh-Kaan.", mithgrap, crossbow, lightsource);
 		dorgStairsChest = new ObjectStep(this, ObjectID.DORGESH_1STAIRS_POSH, new WorldPoint(2721, 5360, 0),
 			"Climb the stairs to the second level of Dorgesh-Kaan.", lockpick);
 
+		dorgLadderToAgi = new ObjectStep(this, ObjectID.DORGESH_CAVEWALL_SLOPE_STEPS, new WorldPoint(2716, 5241, 0),
```
Zoinkwiz, line 215: I think rather than having this be its own step in the sidebar, instead it can be a substep of a comparable step in the sidebar. For example, perhaps we make `moveToUndergroundPylon` say something like `Travel to the Dorgesh-Kaan Agility Course. Either go there through the Lumbridge Castle kitchen's basement, or use a fairy ring to AJQ.`, and make this a sidebar substep?

Additionally, I think the text for it doesn't need to include the condition of 'if teleported to fairy ring', as inherently if you're there all you care about is where to go next, not how you got there.

henrywild: That makes sense, I'll make the changes later

henrywild: @Zoinkwiz Should be resolved with new commit

PR #2685, opened 2026-04-18
Update manual save to be per-user

Zoinkwiz:


src/main/java/com/questhelper/panel/ManualStepSkipStore.java
pajlada: do you need to check for `cm.getRSProfileKey() == null` here too?

pajlada (approved):
Haven't tested this, but it looks sane.
If I understand it correctly, it will mean people have to tick off some stuff again right?

Zoinkwiz:
This will have a translation from existing implementation to the current user's saved config. So there's a scenario that the player for the first time after this update load up this helper on an account they don't want the state on, but I'm hoping most people would have their next load be on the same account they've been using so this won't be noticed.

pajlada:
oh yeah, makes sense. missed the entire legacy loading part https://github.com/Zoinkwiz/quest-helper/pull/2685/changes#diff-7f20a4a434f04c965cb5aff9b7db4f5f319bdde34e02bedbadce47ac6dff4542R52-R61 :sunglasses:

PR #2686, opened 2026-04-18
Leagues: fix glass-making dupe completion and simplify manual skip wiring

blakephillips:
Add QuestStep#withPersistedManualSkip(key) that attaches a ManualRequirement and its persistence key to the step itself, and replace the standalone orderManual_* fields in FauxLeaguesHelper with calls to it. ConditionalStep conditions now read each step's requirement via step.getSidebarManualSkipRequirement() instead of a parallel field.

Fix the section 1 glass-making steps (talktotradercrewmember2, usefurnace2, depositmoltenglass) that all latched on the first trip because their passOnceCompleted conditions were inventory-snapshot equivalents of the first trip's. They now gate on bank totals so the second trip is actually required:
- depositmoltenglass: 10 banked and none in inventory
- talktotradercrewmember2: 10 sand + 10 soda ash + 10 molten glass banked
- usefurnace2: 20 molten glass banked

Also check shears/glassblowingPipe in bank for tradeartima, if they've already been purchased.

Tested changes locally, and verified the manual overrides are backwards compatible and we don't lose any manual requirement overrides.


Zoinkwiz (approved):
Looks good, the cleanup for the manualreqs is great too. Appreciate it!

PR #2688, opened 2026-04-18
Swapped passOnceCompleted with baked in method for ConditionalStep

Zoinkwiz:
This makes it far more succinct to make use sidebar manual steps in helpers.


blakephillips (approved):
Nice work!

PR #2730, opened 2026-05-18
feat(api): expose public QuestHelperApi for inter-plugin integration

Bobakanoosh:
Adds a stable surface that other RuneLite plugins can use to:
- list the quests known to Quest Helper (QuestHelperApi#getQuests)
- open the side panel and select a specific quest (openQuest)
- open the side panel without changing selection (openPanel)

The QuestHelperPlugin class now implements QuestHelperApi, and the
interface + QuestSummary DTO only reference JDK types so the API
remains callable reflectively from plugin-hub plugins that do not
share a classloader.

I plan to use this in https://github.com/Bobakanoosh/runelite-bruhsailor

### DIFF
--- a/src/main/java/com/questhelper/QuestHelperPlugin.java
+++ b/src/main/java/com/questhelper/QuestHelperPlugin.java
@@ -598,12 +598,6 @@ public class QuestHelperPlugin extends Plugin implements QuestHelperApi
 		});
 	}
 
-	// --- QuestHelperApi ------------------------------------------------------
-	//
-	// Stable surface for other RuneLite plugins. See com.questhelper.api for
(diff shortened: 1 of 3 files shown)


pajlada:
I don't think you know what you're doing here - you are not able to use this "api" in a runelite plugin.

Bobakanoosh:
@pajlada Is it possible to accomplish what I'm aiming for, or is there a restriction on cross plugin communication?

pajlada:
It is possible yes just not like this

pajlada:
Well, some of it is possible. See the shortest path plugin and how we interact with it

PR #2731, opened 2026-05-21
Implement The Red Reef helper

### LINKED ISSUE #2734: Missing Quest: The Red Reef

pajlada:
This implements the helper for the quest "The Red Reef"

 - [ ] Is the difficulty correct?
 - [ ] Are the skill requirements correct?
 - [ ] Is the position in the optimal quest guide correct?
 - [ ] Is the position in the ironman optimal quest guide correct?
 - [ ] Is the position in the release order correct?
 - [ ] Are the monsters you have to fight correctly listed?
 - [ ] Are the rewards correct?
 - [ ] Does the "RuneScape Wiki" link work?

A few quirks: The highlight to board at Last Light does not work because you're in an instance & the object is on a different worldview.
I do not attempt to highlight the Zenith boat when you're prompted to board it.
I am not sure if you're able to repair the coral dredgers in a different order, and if you can, if that messes things up. It was a mess to test so it might have some issues.


claudiodekker:
Played through it, and didn't spot anything out of place or broken. 

Regarding the coral dredgers; I only had to repair the northern one, and then run to the eastern one to talk to the NPC, so what was in the helper worked perfectly.

Highlighting the Zenith didn't seem like a problem, but it was also there on my world already (didn't have to hop). It's clearly visible in-game, and the helper's instructions are pretty clear on what to do despite the non-highlighting (the sea arrow re; the direction was good enough for me)

PR #2747, opened 2026-06-22
minimap: use new high precision camera yaw

### LINKED ISSUE #2746: The blue arrow is turning with the compass
Today 6/22/2026 the blue arrow on the minimap is turning with the minimap, so if the compass is not pointing north the arrow will not be pointing to the objective currently.

backwardspy:
fixes #2746

broken by https://github.com/runelite/runelite/commit/ce5918d33a3cb5dd7f9261e1797f4a4334d23fff#diff-bd72e0749a6b55bec42902046c682dcb4390d2fd9918a038e74e2c1335338252

`getCameraYawTarget()` now uses 14-bit angles with new `SINE14` and `COSINE14` tables to match. this caused the `QuestPerspective` code to get an angle 8x higher than it was expecting.

fixed by using the new mask & tables as per upstream's [`localToMinimap` impl](https://github.com/runelite/runelite/blob/master/runelite-api/src/main/java/net/runelite/api/Perspective.java#L605-L608).

note: i can't build & test this at the moment, sorry about that. just hoping it points in the right direction.

cheers!

### DIFF
--- a/src/main/java/com/questhelper/steps/tools/QuestPerspective.java
+++ b/src/main/java/com/questhelper/steps/tools/QuestPerspective.java
@@ -272,10 +272,10 @@ public class QuestPerspective
 			return null;
 		}
 
-		final int angle = client.getCameraYawTarget() & 0x7FF;
+		final int angle = client.getCameraYawTarget() & 0x3FFF;
 
-		final int sin = Perspective.SINE[angle];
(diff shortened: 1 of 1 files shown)


backwardspy:
oh, you already have a fix in a branch. serves me right for not checking first 😄

PR #2756, opened 2026-07-02
feat: allow other plugins to start any helper via PluginMessage

### LINKED ISSUE #2768: PluginMessage hook to start a helper from another plugin?
Would you be open to a PluginMessage hook so another plugin can start a quest helper? I'm working on a plugin where it'd be handy to send someone straight to the helper for a quest, instead of telling them to go find it in the sidebar.

Thinking something like:

```java
eventBus.post(new PluginMessage("questhelper", "launch", Map.of("quest", "Dragon Slayer II")));
```

Looks like it could just call `QuestMenuHandler.startUpQuest()` so the Shield of Arrav / RFD special cases keep working the same as the right click menu does. Happy to PR it if you're up for it.

ajkatz:
Quest Helper already participates in the RuneLite `PluginMessage` ecosystem *outbound* — it posts to `shortestpath` (DetailedQuestStep) and `notenoughrunes` (QuestRequirementsPanel) — but has no inbound receiver, so other plugins can't deep-link into it. This adds the other direction using the same convention:

| field | value |
|---|---|
| **namespace** | `questhelper` |
| **name** | `start` |
| **data** | `{ "quest": "<helper display name>", "source": "<requesting plugin display name>" }` — both required |

```java
eventBus.post(new PluginMessage("questhelper", "start",
    Map.of("quest", "Dragon Slayer I", "source", "<My Plugin>")));
eventBus.post(new PluginMessage("questhelper", "start",
    Map.of("quest", "Ardougne Elite Diary", "source", "<My Plugin>")));
eventBus.post(new PluginMessage("questhelper", "start",
    Map.of("quest", "Herb run", "source", "<My Plugin>")));      // generic helper
eventBus.post(new PluginMessage("questhelper", "start",
    Map.of("quest", "Agility", "source", "<My Plugin>")));       // skill helper
```

**The start is announced in the player's chat** — `Quest Helper for Dragon Slayer I started, requested by My Plugin.` — so it's always visible which plugin triggered it. Messages without a `source` are ignored.

**Works for every helper type**, not just quests: the handler delegates to the existing `QuestMenuHandler#startUpQuest(String)` — the same generic path the quest list and `onChatMessage` auto-start use — so quests, miniquests, **achievement diaries, skill helpers, generic helpers, and player-made quests** all resolve through `QuestHelperQuest.getByName`, and the Shield of Arrav / RFD special cases are inherited rather than duplicated. The call is wrapped in `clientThread.invokeLater` since `PluginMessage` arrives on the poster's thread and the Shield of Arrav branch reads the local player.

Malformed or foreign messages are ignored (namespace/name gate + payload type checks), so nothing changes for anyone who doesn't post the message.

Motivation: I maintain the Goal Planner hub plugin and want "resume in Quest Helper" on quest **and diary** goals — but this is deliberately generic so any plugin (or future Quest Helper features) can use it. Tested end-to-end: a dev client running both plugins, with Goal Planner posting the message and this handler starting quest and diary helpers.

`./gradlew build` green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/java/com/questhelper/QuestHelperPlugin.java
+++ b/src/main/java/com/questhelper/QuestHelperPlugin.java
@@ -53,7 +53,9 @@ import net.runelite.api.gameval.VarPlayerID;
 import net.runelite.api.gameval.VarbitID;
 import net.runelite.client.RuneLite;
 import net.runelite.client.callback.ClientThread;
+import net.runelite.api.ChatMessageType;
 import net.runelite.client.chat.ChatMessageManager;
+import net.runelite.client.chat.QueuedMessage;
 import net.runelite.client.config.ConfigManager;
 import net.runelite.client.eventbus.EventBus;
 import net.runelite.client.eventbus.Subscribe;
@@ -611,9 +613,14 @@ public class QuestHelperPlugin extends Plugin
 	 * quests. Post a message with namespace "questhelper", name "start", and
 	 * a "quest" entry in the data map containing the helper's display name:
 	 * </p>
+	 * The data map must also carry a "source" entry with the requesting
+	 * plugin's display name; the start is announced in the player's chat so
+	 * it is always visible which plugin triggered it.
 	 * <pre>{@code
-	 * eventBus.post(new PluginMessage("questhelper", "start", Map.of("quest", "Dragon Slayer I")));
-	 * eventBus.post(new PluginMessage("questhelper", "start", Map.of("quest", "Ardougne Elite Diary")));
+	 * eventBus.post(new PluginMessage("questhelper", "start",
+	 * 	Map.of("quest", "Dragon Slayer I", "source", "My Plugin")));
+	 * eventBus.post(new PluginMessage("questhelper", "start",
+	 * 	Map.of("quest", "Ardougne Elite Diary", "source", "My Plugin")));
 	 * }</pre>
 	 *
 	 * @param event The plugin message posted by another plugin.
@@ -627,13 +634,21 @@ public class QuestHelperPlugin extends Plugin
 		}
 
 		Object quest = event.getData().get("quest");
-		if (!(quest instanceof String))
+		Object source = event.getData().get("source");
+		if (!(quest instanceof String) || !(source instanceof String))
 		{
-			log.debug("Ignoring quest helper start plugin message with invalid quest payload: {}", quest);
+			log.debug("Ignoring quest helper start plugin message with invalid quest or source payload: {} (source {})", quest, source);
 			return;
 		}
 
-		clientThread.invokeLater(() -> questMenuHandler.startUpQuest((String) quest));
+		clientThread.invokeLater(() ->
+		{
+			chatMessageManager.queue(QueuedMessage.builder()
+				.type(ChatMessageType.GAMEMESSAGE)
+				.runeLiteFormattedMessage("Quest Helper for " + quest + " started, requested by " + source + ".")
+				.build());
+			questMenuHandler.startUpQuest((String) quest);
+		});
 	}
 
 	public void displayPanel()
--- a/src/main/java/com/questhelper/QuestHelperPlugin.java
+++ b/src/main/java/com/questhelper/QuestHelperPlugin.java
@@ -618,9 +618,13 @@ public class QuestHelperPlugin extends Plugin
 	 * it is always visible which plugin triggered it.
 	 * <pre>{@code
 	 * eventBus.post(new PluginMessage("questhelper", "start",
-	 * 	Map.of("quest", "Dragon Slayer I", "source", "My Plugin")));
+	 * 	Map.of("quest", "Dragon Slayer I", "source", "<My Plugin>")));
 	 * eventBus.post(new PluginMessage("questhelper", "start",
-	 * 	Map.of("quest", "Ardougne Elite Diary", "source", "My Plugin")));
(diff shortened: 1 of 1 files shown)


src/main/java/com/questhelper/QuestHelperPlugin.java
```diff
@@ -599,6 +600,39 @@ public void onChatMessage(ChatMessage chatMessage)
...
+	 * display name:
+	 * </p>
+	 * <pre>{@code
+	 * eventBus.post(new PluginMessage("questhelper", "start", Map.of("quest", "Dragon Slayer I")));
+	 * }</pre>
+	 *
+	 * @param event The plugin message posted by another plugin.
+	 */
+	@Subscribe
+	public void onPluginMessage(PluginMessage event)
```
pajlada, line 633: I'm not opposed to this functionality, however I would like the map to be required to contain the source plugin display name so we can send a chat message, something like: "Quest Helper for quest Dragon Slayer I started, requested by Plugin Name"

pajlada:
The snapshot build failures are fine, they're an issue on our end not yours

ajkatz:
> The snapshot build failures are fine, they're an issue on our end not yours

thank you for the quick feedback! I'll wait until those are fixed

ajkatz:
@pajlada rebased this and looks like it's no longer hitting that build issue now 🥳

VinnyVanGogh:
A complementary question rather than a competing proposal — I don't want to duplicate work here.

This gives plugins a way to **set** Quest Helper's selection, but there's still no way to **read** it. `QuestHelperPlugin#getSelectedQuest()` is public, but a hub plugin can't call it: `QuestHelper` and `QuestHelperQuest` are Quest Helper's own types in Quest Helper's classloader, so reaching them means reflection, which is an automatic Plugin Hub rejection.

Concrete case: a plugin that tracks quest progress and wants its panel to follow whatever Quest Helper currently has open. With this PR it can push a selection, but it can't tell that the user then picked something else from the sidebar, so the two views silently drift apart.

The smallest thing that closes it, staying inside the protocol you've defined:

- **Outbound**, whenever the selection changes: `PluginMessage("questhelper", "selected", Map.of("quest", <helper display name>))` — and the same message with an empty data map when the selection is cleared. One name rather than a `selected`/`cleared` pair, so a subscriber has a single code path: `data.get("quest") instanceof String`, otherwise nothing is selected.
- **Inbound**, so a plugin that starts mid-session isn't stuck until the next change: `PluginMessage("questhelper", "getSelected", Map.of("source", "<My Plugin>"))`, answered by posting that same `selected` message. `source` required and validated the same way `start` validates it. No chat announcement — reading isn't a user-visible action, and announcing a poll would spam the chatbox.

The emit points are `QuestManager#initializeNewQuest` and `QuestManager#shutDownQuest`, both of which already run on the client thread.

**Why I'm asking before writing it:** RuneLite's `EventBus` requires a subscriber to be named `on<EventName>`, so a second `PluginMessage` handler in `QuestHelperPlugin` would need the `@Subscribe(preferredName = ...)` escape hatch, and `QuestManager` isn't registered on the bus at all. The natural shape is a single `onPluginMessage` switching on `event.getName()` — which is exactly the method this PR adds. So a separate PR conflicts with yours by construction.

Options, in the order I'd guess you'd prefer them:

1. **Fold it into this PR.** I'll hand you a patch, or open a PR against `ajkatz:plugin-message-start` for you to merge in.
2. **Separate PR** that I rebase on top once this one merges.
3. **Not wanted** — that's a fine answer too, and better to hear it now.

Happy to go whichever way you two prefer. Related: #2768.

pajlada:
@VinnyVanGogh please write a human sentence. I'm not reading your two-pager.

VinnyVanGogh:
Sorry, that was way too long. I was trying to answer every follow-up up front so you wouldn't have to ask, which obviously backfired.

Short version: I can't find any way for another plugin to read which helper is currently selected. This PR lets you set it, but there's no way to read it back, and reflection is a Hub rejection so there's no way around that.

I have a working patch that I could add to this PR, or keep separate if you'd rather.

pajlada:
> Sorry, that was way too long. I was trying to answer every follow-up up front so you wouldn't have to ask, which obviously backfired.
> 
> Short version: I can't find any way for another plugin to read which helper is currently selected. This PR lets you set it, but there's no way to read it back, and reflection is a Hub rejection so there's no way around that.
> 
> I have a working patch that I could add to this PR, or keep separate if you'd rather.

I think an appropriate way to handle this would be QuestHelper sending a PluginMessage whenever the user selects a quest. Then you could, in your plugin, listen to PluginMessage's on QuestHelper's outgoing namespace and act on that.
Does that sound reasonable?

VinnyVanGogh:
> > Sorry, that was way too long. I was trying to answer every follow-up up front so you wouldn't have to ask, which obviously backfired.
> > Short version: I can't find any way for another plugin to read which helper is currently selected. This PR lets you set it, but there's no way to read it back, and reflection is a Hub rejection so there's no way around that.
> > I have a working patch that I could add to this PR, or keep separate if you'd rather.
> 
> I think an appropriate way to handle this would be QuestHelper sending a PluginMessage whenever the user selects a quest. Then you could, in your plugin, listen to PluginMessage's on QuestHelper's outgoing namespace and act on that. Does that sound reasonable?

Yeah, I think that would cover it. Do you want me to work on that?

PR #2765, opened 2026-07-19
Add PluginMessage state bridge for Quest Journal

nanopink:
Quest Journal is a separate Plugin Hub UI and cannot access Quest Helper classes because Plugin Hub plugins use isolated classloaders.

This adds a versioned `PluginMessage` producer for the quest list, progress totals, selected quest overview, objectives, requirements, recommendations, enemies, rewards, and list defaults. Payloads contain only strings, numbers, booleans, lists, and maps. The producer activates only after a consumer request and republishes only changed state.

Quest starting is intentionally out of scope here and left to the source-attributed protocol proposed in #2756, avoiding a conflicting silent selection path.

The patch also exposes item reward ID and quantity so consumers can render accurate reward icons.

Tests: `./gradlew test --rerun-tasks` on JDK 21 passes all 118 tests, including 23 bridge tests.


nanopink:
This PR will be closed to attempt to minimize the code on your side by moving some parsing logic to my side, so it's easier for your maintainers to evaluate and merge it.

PR #2773, opened 2026-07-29
feat: Added Fallen From Grace

Zoinkwiz:


src/main/java/com/questhelper/helpers/quests/fallenfromgrace/FallenFromGrace.java
```diff
@@ -0,0 +1,447 @@
...
+	}
+
+
+	void setupSteps()
+	{
+		travelToWyrmscraig = new SailStep(this, Port.WYRMSCRAIG);
+
+		talkToFfion = new NpcStep(this, NpcID.WYRMSCRAIG_VILLAGER_1, new WorldPoint(2575, 2296, 0), "Talk to Ffion to be allowed onto the island.");
+
+		talkToCormac = new NpcStep(this, NpcID.CORMAC, new WorldPoint(2576, 2255, 0), "Talk to Cormac in Auchrie village on Wyrmscraig island.");
```
Daamsed, line 193: There is a shortcut you can take prior to this step. Should this shortcut be highlighted?

src/main/java/com/questhelper/helpers/quests/fallenfromgrace/FallenFromGrace.java
```diff
@@ -0,0 +1,447 @@
...
+
+		killBossSidebar = new DetailedQuestStep(this, "Defeat the Mad Angel. Use Protect from Melee.");
+		killBossSidebar.addText("If she pulls her sword back slowly, run behind her.");
+		killBossSidebar.addText("If she charges a blue attack, protect from magic. If you activate it the tick it hits you, you get a guaranteed max hit on her next hit.");
+		killBossSidebar.addText("If she launches a projectile at a tile, stand on the tile to bounce it back to her.");
+		killBossSidebar.addSubSteps(killBoss);
+
+		searchBody = new ObjectStep(this, ObjectID.FFG_CORPSE_KEENAN, new WorldPoint(2529, 2215, 0), "Search the body in the church.");
+		readNote = new DetailedQuestStep(this, "Read the final letter.", finalLetter.highlighted());
+		finishQuest = new NpcStep(this, NpcID.CORMAC, new WorldPoint(2576, 2255, 0), "Return to Cormac to finish the quest.");
```
Daamsed, line 254: Maybe worth mentioning prior to this step that you can leave via the Cathedral door (requires the user to pick lock), and then via the shortcut that needs to be cleared?

Daamsed (approved):
I completed the quest using this branch, and I didn't experience any real issues. Did not inspect the code too much.

build.gradle
```diff
@@ -11,7 +11,7 @@ repositories {
 }
 
 dependencies {
-	def runeLiteVersion = "latest." + (project.hasProperty("use.snapshot") ? "integration" : "release")
+	def runeLiteVersion = "latest.integration"
```
pajlada: Remember to undo before merging in

pajlada (changes requested):
Noticed a few things that I wrapped up in this PR https://github.com/Zoinkwiz/quest-helper/pull/2775 - feel free to make the changes manually instead of merging the PR.

1) The sidebar would get lost when unlocking the chest with the combination lock
2) Gameval'd the "is combination lock open" check, and using a widget presence check instead of widget text comparisons
3) added the quest to the list of opted in "all sidebars are accounted for" tests

PR #2787, opened 2026-08-15
polish: remove unnecessary step from shilo village

aivot-on:
Tested these recently and added to the wikis quick guide.


pajlada (approved):
This looks reasonable if the step isn't required 👍

aivot-on:
> This looks reasonable if the step isn't required 👍

Yeah I wish I would've checked the quest helper code before spending whole yesterday figuring out the minimal steps based on the wiki guides and realizing the comments in the guide code. Oh well, one unnecessary step hopefully removed as a result.

Zoinkwiz (approved):

PR #2808, opened 2026-08-27
Ethically Acquired Antiquities: highlight crewmembers and make charter coins optional

nwheise:
Two fixes to the Ethically Acquired Antiquities helper, both around the Port Sarim step.

Screenshot 2026-08-26 at 9 25 52 PM

### Trader Crewmembers are valid targets, but weren't highlighted

The [wiki](https://oldschool.runescape.wiki/w/Ethically_Acquired_Antiquities) says to talk to "Trader Stan or any of the Trader Crewmembers who are located at the most south dock", and either one progresses the quest. The step only pointed at Trader Stan.

- Reworded the step to "speak to Trader Stan or any Trader Crewmember on the southern deck"
- Added `getCrewMembers()` to the step's alternate NPCs — the same list the two Fortis Cothon crewmember steps already use
- Set `allowMultipleHighlights`, since `NpcStep` otherwise tracks only the first matching NPC, which meant Stan alone was highlighted even once the crew IDs were valid targets

The dialog options were already wired for either NPC and are unchanged.

### 3000 coins is optional, not required

The coins are only needed if you charter a ship to Port Sarim; the wiki notes you can travel "by any method you wish, such as the charter ship, the Rat Pits Minigame Teleport, or a planted spirit tree". They were listed as a hard requirement in three places, so the quest showed "3000 x Coins" as required for everyone.

- `getItemRequirements()` → empty; coins moved to `getItemRecommended()`
- Step uses `addRecommended()` rather than a constructor requirement, so the overlay lists them under "Recommended:"
- Section panel uses the 4-arg `PanelDetails` constructor to pass them as recommended
- Added a tooltip explaining they're only needed for the charter

### Testing

Added `EthicallyAcquiredAntiquitiesTest` covering both behaviors — that the step's `allIds()` includes both Trader Stan and the Port Sarim crewmembers, and that the coins are recommended rather than required. Both fail without the corresponding fix.

`./gradlew build` passes (97 tests). Also verified in-game at the Port Sarim dock: the crewmembers and Stan all highlight, and the coins render under "Recommended:".


src/main/java/com/questhelper/helpers/quests/ethicallyacquiredantiquities/EthicallyAcquiredAntiquities.java
```diff
@@ -259,13 +263,13 @@ private List<Integer> getCrewMembers()
 	@Override
 	public List<ItemRequirement> getItemRequirements()
 	{
-		return Arrays.asList(coinsForCharter);
+		return Collections.emptyList();
```
pajlada: The coins should still be in the list of requirements, as long as there's a highlight for it. Remove it from recommendations after re-adding it here.

src/main/java/com/questhelper/helpers/quests/ethicallyacquiredantiquities/EthicallyAcquiredAntiquities.java
```diff
@@ -180,12 +181,15 @@ public void setupSteps()
 		((NpcStep) returnToCrewmember).addAlternateNpcs(getCrewMembers());
 
 		returnToCrewmember.addDialogStep("So, about that man with the case.");
-		talkToTraderStan = new NpcStep(this, NpcID.SAILING_TRANSPORT_TRADER_STAN_BASE, new WorldPoint(3036, 3194, 0), "Head to Port Sarim and speak to Trader Stan on the southern " +
-				"deck. Charter at a cost of 3000 coins.", coinsForCharter);
+		talkToTraderStan = new NpcStep(this, NpcID.SAILING_TRANSPORT_TRADER_STAN_BASE, new WorldPoint(3036, 3194, 0), "Head to Port Sarim and speak to Trader Stan or any Trader " +
+				"Crewmember on the southern deck. You can charter a ship from Civitas illa Fortis to Port Sarim for 3000 coins.", true);
+		((NpcStep) talkToTraderStan).addRecommended(coinsForCharter);
```
pajlada, line 186: At times like these, we purposefully don't highlight all options. If there's a solid option, we only highlight that. There's no need to give a bunch of alternatives right next to the option we've already given.

src/main/java/com/questhelper/helpers/quests/ethicallyacquiredantiquities/EthicallyAcquiredAntiquities.java
```diff
@@ -303,7 +307,7 @@ public List<PanelDetails> getPanels() {
 		List<PanelDetails> allSteps = new ArrayList<>();
 		allSteps.add(new PanelDetails("Grand Museum", Arrays.asList(inspectEmptyDisplayCase, talkToCuratorHerminius, investigateToolsBehindDisplayCase, investigateCaseAgain, talkToAcademic, talkToTourist, talkToCitizen), Arrays.asList(civitasIllaFortisTeleport)));
 
-		allSteps.add(new PanelDetails("Done with museum visit", Arrays.asList(talkToRegulus, talkToCrewmember, talkToArtima, returnToCrewmember, talkToTraderStan, talkToBetty, readBettysNotes), Arrays.asList(coinsForCharter)));
+		allSteps.add(new PanelDetails("Done with museum visit", Arrays.asList(talkToRegulus, talkToCrewmember, talkToArtima, returnToCrewmember, talkToTraderStan, talkToBetty, readBettysNotes), Collections.emptyList(), Arrays.asList(coinsForCharter)));
```
pajlada: Same here - add the requirement back, as long as it keeps the note you made that it's not required if you're not charting.

src/test/java/com/questhelper/helpers/quests/ethicallyacquiredantiquities/EthicallyAcquiredAntiquitiesTest.java
```diff
@@ -0,0 +1,49 @@
...
+	@Test
+	void ensureCharterCoinsAreRecommendedRatherThanRequired()
+	{
+		var helper = setupHelper();
+
+		assertFalse(helper.getItemRequirements().contains(helper.coinsForCharter),
+			"Coins are only needed if you charter a ship to Port Sarim, so they must not be required");
+		assertTrue(helper.getItemRecommended().contains(helper.coinsForCharter),
+			"Coins should be recommended for the charter ship to Port Sarim");
+	}
```
pajlada: These unit tests don't really serve a purpose

pajlada (changes requested):

src/test/java/com/questhelper/helpers/quests/ethicallyacquiredantiquities/EthicallyAcquiredAntiquitiesTest.java
nwheise: sg, I will remove

src/main/java/com/questhelper/helpers/quests/ethicallyacquiredantiquities/EthicallyAcquiredAntiquities.java
nwheise: see other comment, I'll remove if you confirm that's your intent

src/main/java/com/questhelper/helpers/quests/ethicallyacquiredantiquities/EthicallyAcquiredAntiquities.java
nwheise: Gotcha - fwiw, I just did this quest and was just following the quest helper to do it quickly. This step currently says "Head to Port Sarim and speak to Trader Stan on the southern deck. Charter at a cost of 3000 coins.", and only highlights Trader Stan.

Here's what I did at this step:
1. Quest helper says Head to Port Sarim, so I did without chartering
2. Quest helper says talk to Trader Stan. **His correct dialog option here is buried under "More Options", so no dialog option is highlighted blue**
3. Quest helper says "Charter a ship at 3000 coins", so I almost chartered a ship back to Civitas illa Fortis before catching myself.

I went to the wiki to figure out what I actually needed. So this confusion is what I'm trying to fix with this PR

src/main/java/com/questhelper/helpers/quests/ethicallyacquiredantiquities/EthicallyAcquiredAntiquities.java
nwheise: Thanks for the comments - by highlight, do you mean a talk dialog highlight? There is not one, the coins here are totally unused if you travel to Port Sarim a different way

pajlada: sorry, by highlight i meant to type tooltip (which is what you added in this PR)

src/main/java/com/questhelper/helpers/quests/ethicallyacquiredantiquities/EthicallyAcquiredAntiquities.java
pajlada: I think the correct fix in this case would be to correctly highlight the dialog options necessary to get to the correct dialog option. The change to the "Charter a ship" step is reasonable

nwheise:
Here's how it looks:
Screenshot 2026-08-27 at 9 40 36 AM
* 3000 coins in recommended items, since it's a fast-travel option, like the teleports
* Hover-over tooltip still works
* Only one NPC highlighted
* More clarity in task description

Screenshot 2026-08-27 at 9 41 17 AM

src/main/java/com/questhelper/helpers/quests/ethicallyacquiredantiquities/EthicallyAcquiredAntiquities.java
nwheise: agreed, unfortunately I don't know how to re-test with More Options dialog now that I've completed the quest. So I'll just leave it

src/main/java/com/questhelper/helpers/quests/ethicallyacquiredantiquities/EthicallyAcquiredAntiquities.java
```diff
@@ -259,13 +263,13 @@ private List<Integer> getCrewMembers()
 	@Override
 	public List<ItemRequirement> getItemRequirements()
 	{
-		return Arrays.asList(coinsForCharter);
+		return Collections.emptyList();
 	}
```
pajlada: If you don't think we should have any item requirements, you can remove the entire function

nwheise: removed

pajlada (approved):

PR #2817, opened 2026-09-09
Tree Run Update

maskarauder:
- Adds requirements for planting calquats.
- Adds None option for each tree type that disables those patches.
- 
> [!IMPORTANT]
> This PR depends on #2823. Please merge #2823 before merging this PR.


maskarauder:
Ready for testing.

Made some changes to how visibility propagates out of nested panels, so want to be sure nothing breaks.

Will need to test the following to be sure they still work:
1. Achievement diary
2. Charting (entering and leaving Done for now)
3. Herb run (I can do an overhaul later)
4. DT2

pajlada:
@maskarauder Could you split up the panel visibility changes into a separate PR? It would be simplest to get that tested and merged in first

maskarauder:
Pushed to remove the panel changes from this commit since they're covered in #2823 

This PR now depends on #2823 getting merged first.

PR #2823, opened 2026-09-12
Fix hide conditions for outer panels when nested panels are updated.

maskarauder:
Prevents outer panels from showing up in the sidebar when all inner panels are hidden.


src/main/java/com/questhelper/panel/queststepsection/QuestSectionSection.java
```diff
@@ -247,19 +247,28 @@ public void mouseExited(MouseEvent mouseEvent)
 
 	public boolean updateStepVisibility(Client client)
 	{
-		boolean stepVisibilityChanged = false;
+		boolean visibilityChanged = false;
```
pajlada: You could undo this variable name change since it's still checks for the step visibility

pajlada (approved):
In my testing, this seems to work fine. Just a refactor nit if you can be bothered.

maskarauder:
I changed the variable names back.

PR #2827, opened 2026-09-14
feat: allow excluding patch types from the Tree Run

Daaam4:
## Summary

Adds a **None** option to each of the sapling dropdowns in the Tree Run configuration (Trees, Fruit Trees, Hardwood Trees, Calquat Trees). Selecting it excludes that patch type from the run entirely — e.g. if you don't want to bother with calquats, set **Calquat Trees → None** and the Tai Bwo Wannai / Kastori calquat / Great Conch patches disappear from the run.

## What changes when a patch type is set to None

- **Step routing** – the patches are skipped when the helper picks the next step (all `steps.addStep(...)` conditions are gated on `not(<type>Excluded)`), including the sub-steps inside grouped regions (Farming Guild, Gnome Stronghold, Karamja, Fossil Island, Kastori).
- **Sidebar** – the panels for those patches are hidden via `PanelDetails#withHideCondition`. Grouped panels (e.g. *Kastori* with fruit + calquat) only hide when *all* their children are excluded.
- **Item requirements** – the sapling, protection payment and 200gp clearing coins for those patches are no longer counted / shown, so the bank tab and requirements list stay accurate.
- **"Wait for trees"** – excluded patch types count as done, so they don't block the wait panel.

## Implementation notes

- `FarmingUtils.PlantableItem` gains `NONE_ITEM_ID` (`-1`), an `isNone()` helper, and the default `get*ItemRequirement` methods return a placeholder instead of calling `ItemManager#getItemComposition(-1)`.
- `NONE` is appended to `TreeSapling`, `FruitTreeSapling`, `HardwoodTreeSapling` and `CalquatTreeSapling`. Existing config values are unaffected (defaults unchanged).
- `TreeRun` exposes four `RuneliteRequirement`s (`treesExcluded`, `fruitTreesExcluded`, `hardwoodTreesExcluded`, `calquatTreesExcluded`) that key off the existing config entries, so no new config keys are introduced.
- The four near-identical `update*Sapling` / `update*PaymentItem` methods were collapsed into `updateSaplingItems` / `updatePaymentItems` which handle the `NONE` case.

## Testing

- `./gradlew compileJava test` passes.
- Not yet verified in-client — would appreciate a quick sanity check from someone with a RuneLite dev setup handy.


pajlada:
This is a duplicate of https://github.com/Zoinkwiz/quest-helper/pull/2817

PR #2831, opened 2026-09-16
PrinceAliRescue note on rope price

### LINKED ISSUE #2816: Ok this is very minor but set me back a minute... Price Ali Rescue quest...
says rope is 15gp but is 18gp. less is more bois, or is it

vmgi:
Added a note mentioning the rope can be bought for 18 coins from Ned's shop as opposed to 15 coins over dialogue. For those who may be confused to find the 18gp price tag before running through the dialogue.
resolves #2816


vmgi:
Resolves #2816

pajlada (approved):

PR #2832, opened 2026-09-16
TempleOfTheEye fix tea step condition

### LINKED ISSUE #2803: Temple of the eye: Getting the Mage of Zamorak tea.
After you get the strong tea from the tea seller, if you don't have a bucket of water or the eye amulet on you, the step still says to talk to the tea seller, when you should be getting a bucket of water and amulet and doing the next step of talking to the Mage of Zamorak.

vmgi:
step does not require the removed items, these items are listed in the next step where they are actually needed. 
resolves #2803


pajlada:
This is testable on a speedrunning world to make sure it works as expected

pajlada (approved):
I tested this on a speedrunning world and this works as expected 👍

vmgi:
> This is testable on a speedrunning world to make sure it works as expected

Awesome, thanks for the insight. It's been a min 😄

PR #2835, opened 2026-09-17
Let another plugin start a quest via PluginMessage

AltDoug:
Hey, I'm Doug (AltDoug). This is a small addition, not something I'm precious about, so feel free to change the namespace, event name, or approach to whatever fits your codebase better.

**What it does**

Adds an inbound `PluginMessage` listener on `QuestManager` so another RuneLite plugin can ask Quest Helper to start a named quest, without reflecting into Quest Helper's internal classes to do it.

```java
eventBus.post(new PluginMessage("questhelper", "start", Map.of("quest", "Cook's Assistant")));
```

The namespace (`questhelper`) and event name (`start`) mirror the shape Quest Helper already uses for its own outbound message to Shortest Path (`DetailedQuestStep#setShortestPath()` posts `new PluginMessage("shortestpath", "path", data)`). Both are yours to rename if you'd prefer different ones.

One other line worth flagging: this also registers `QuestManager` on the event bus in `QuestHelperPlugin#startUp` (and unregisters it in `shutDown`), since it wasn't registered before and the listener wouldn't receive anything without it. `onPluginMessage` is the only `@Subscribe` in `QuestManager`, so nothing else starts firing as a result of that registration.

**What it doesn't do**

Nothing changes for a player unless another plugin actually posts the message. It's additive: a message from any other namespace, or one with a missing or unresolvable "quest" value, is ignored, with no exception and no side effect.

**Why I'm asking for this instead of reflecting**

I maintain a plugin that hands a specific quest off to Quest Helper by name. Right now the only way to do that from outside Quest Helper is reflection into its internal classes, which breaks whenever Quest Helper refactors internally, and is probably not something you want third-party plugins relying on anyway. A supported inbound message removes that need.

**Testing**

Added `QuestManagerPluginMessageTest`, covering: a message in the `questhelper` namespace with a resolvable quest name starts that quest, a message in another namespace is ignored, a message with a missing or unresolvable quest name is ignored without throwing, and posting the same start message twice resolves to starting the same quest both times. `./gradlew build` passes locally.

Thanks for maintaining Quest Helper. It's a solid reference for how a guidance plugin should behave.


pajlada:
Hi, please look through existing pull requests. https://github.com/Zoinkwiz/quest-helper/pull/2756 should already cover this. If you have concerns with that implementation, please comment on that pull request.

